PR #8, opened 2020-03-19
Fixes for FStarLang/FStar#1905

mtzguido:
This PR fixes spurious uses of names, as described in FStarLang/FStar#1905. F* will reject these names with a syntax error once the fix is merged (waiting on this PR and one in miTLS).


mtzguido:
F* now just warns for these things, so this PR is just to remove the warnings. (So if this conflicts with #9, I think you should favour that PR.)

mtzguido:
Not really needed, and a bit stale by now.

PR #9, opened 2020-04-08
Removing abstract from the codebase

aseemr:
This PR removes the abstract keyword from the code.

I have introduced some 10 odd interfaces. In some 4-5 instances I had to resort to `opaque_to_smt` to get around the proofs that were relying on the leaky abstraction.

There are some 4 admits in `LowParse.Low.Base.fsti`, they are small proofs and mostly it requires calling some `equiv f f'` kind of lemmas. @tahina-pro, it might be quite easy for you to do those proofs, can you please take a look, else I can try it some time. Thanks!

I have also enabled `--warn_error @332`.


tahina-pro:
Rechecking LowParse.Low.Base.fsti while checking LowParse.Low.Base.fst fails, so I had to split it into two parts: LowParse.Low.Base.Spec with an interface for abstract definitions only, and LowParse.Low.Base without interface, for everything else, including stateful functions.

tahina-pro:
Thanks a lot Aseem for your work!

PR #23, opened 2020-07-31
Error reporting in terms of 3d file locations and generating a .Types.fst file with just the type definitions

aseemr:
This PR adds support for reporting errors in arithmetic expressions in terms of their 3d locations. It also adds support for generating a `.Types.fst` file that contains just the type definitions (no parsers, actions, validators, etc.).

I have to test it more, but would be good to have some early feedback. Here is how it works right now:

Given a 3d definition as:

```
entrypoint typedef struct _point {
  UINT32 x;
  UINT32 y { y > x + 1 }
} point;
```

We generate a `Point.Types.fst` file, that when typechecked gives an error:

```
(Error) Cannot verify that u32 add is safe(Other related locations: ../prelude/Prelude.fsti(249,30-249,61)); The SMT solver could not prove the query, try to spell your proof in more detail or increase fuel/ifuel
Related location (C-c C-' to visit, M-, to come back): /home/aseem/evm/quackyducky/src/3d/tests/Point.3d(3,18-3,18)
```

The `3d` file location is printed in the related location (can we massage this further in the `everparse.bat` script?).

The verification context for this `Points.Types.fst` is set to:

```
#set-options "--fuel 0 --ifuel 0 --using_facts_from 'Prims FStar Prelude -FStar.Tactics -FStar.Reflection'"
```

Main changes include:

- Adding range to `Target.expr` and copying `Ast.expr.range` when translating
- Adding arithmetic functions in `Prelude.fsti`, e.g.

```
let u16_add (msg:string) (r:Prims.range) (x y:tUInt16t)
  : Pure UInt16.t
      (requires labeled r msg (UInt.size (UInt16.v x + UInt16.v y) UInt16.n))
      (ensures fun z -> UInt16.v z == UInt16.v x + UInt16.v y)
  = UInt16.add x y
```

- Translating arithmetic expressions in `Target` ast to these functions in prelude, e.g. for the example above, the refinement on `y` is translated as:

```
inline_for_extraction noextract
let _point_y_filter (x:(_point_x )) (y:(___UINT32 )) : (bool ) = (y `FStar.UInt32.gt` (x `Prelude.u32_add "Cannot verify that u32 add is safe" (Prims.mk_range "Point.3d" 3 18 3 18)` (FStar.Int.Cast.uint8_to_uint32 1uy)))
```


tahina-pro:
I just added support for verifying the `.Types.fst` files through `3d --batch` (and thus `everparse.bat`/`everparse.sh`), and for cleaning them up through `3d --cleanup`

src/3d/Target.fst
```diff
@@ -82,17 +82,37 @@ let namespace_of_integer_type =
...
+
+let print_arith_op
+  (o:op{Plus? o \/ Minus? o \/ Mul? o \/ Division? o})
+  (t:A.integer_type)
+  (r:option A.range)
+  = let t = match t with | A.UInt8 -> "u8" | A.UInt16 -> "u16" | A.UInt32 -> "u32" | A.UInt64 -> "u64" in
+    let fn = match o with | Plus _ -> "add" | Minus _ -> "sub" | Mul _ -> "mul" | Division _ -> "div" in
+    let r = match r with | Some r -> r | None -> A.dummy_range in
+
+    Printf.sprintf "`Prelude.%s_%s \"Cannot verify that %s %s is safe\" %s`"
```
tahina-pro, line 102: The "Cannot verify that ... is safe" error messages look quite general. Thus, we could add them once and for all in the definitions of `Prelude.u32_mul` etc. instead of generating them in the .Types.fst

tahina-pro:
> The 3d file location is printed in the related location (can we massage this further in the everparse.bat script?).

I think this should be feasible in `ocaml/Batch.ml`

src/3d/Target.fst
aseemr: Done

PR #26, opened 2020-08-06
More arithmetic operators

nikswamy:
Adding support for %, &, |, ^, ~, <<, >>

This PR is on top of @aseemr's error reporting PR.

The main bit I would be happy to have a second set of eyes on is the precedence levels I've assigned for all the operators.


nikswamy:
I also changed boolean negation to use "!" instead of "not" ... which makes it compatible with what was documented.

nikswamy:
Also adds multi-line comments, removes `=` for equality comparisons in favor of `==`.

PR #27, opened 2020-08-30
C3d: Where clauses

mtzguido:
Hi @msprotz, mostly posting this for review. I added support for where clauses mostly mimicking your existing code for parameters. The tricky bit was adding `this` into the parsing context, which I think came out OK (details in commit messages + comments). I haven't yet refactored anything, wanted to check whether the logic seems OK.

For constraints and their scope, I was thinking we could do a similar trick for the scoping: push a new scope before parsing the constraint, push all the previously-parsed fields, and then parse the constraint. That would make the scopes clean, and seems implementable, but would be quadratic--- are we expecting big (~100 field) structures? Ideally we would carry the "ghost" scope while parsing, but seems like that would require clang surgery.

As an example, our good old `basic0.h`:
```cpp
typedef struct
[[
  everparse::entrypoint,
  everparse::process,
  everparse::parameter(uint32_t MessageBodyLength),
  everparse::where(MessageBodyLength == sizeof(this))
]]
_FOO {
  FOO_ID RequestId;
  uint32_t MajorVersion
      [[everparse::constraint(MajorVersion == FOO_MAJOR_VERSION && MajorVersion == 0)]];
  uint32_t MinorVersion
      [[everparse::constraint(MinorVersion == MajorVersion),
        everparse::constraint(MinorVersion == 1)]];
  uint32_t MaxTransferSize
      [[everparse::constraint(MaxTransferSize <= MessageBodyLength)]];
} FOO, *PFOO;
```
Now turns into
```
entrypoint
typedef struct _FOO (uint32_t MessageBodyLength)
where (MessageBodyLength == sizeof (this))
 { 
  FOO_ID RequestId;
  uint32_t MajorVersion { MajorVersion == 0 && MajorVersion == 0 } ;
  uint32_t MinorVersion { (MinorVersion == MajorVersion) && (MinorVersion == 1) } ;
  uint32_t MaxTransferSize { MaxTransferSize <= MessageBodyLength } ;
}
```
The where clause is still interpreted as a string, and printed as-is.

### DIFF
--- a/c3d/src/C3d.cpp
+++ b/c3d/src/C3d.cpp
@@ -685,7 +685,7 @@ public:
             switch (State) {
                 case Mid:
                     Out << " && ";
-                    // fall through!
+                    [[fallthrough]];
 
                 case First:
                     Out << "(" << W->getAnnotation().slice(shift, W->getAnnotation().size()) << ")";
--- a/c3d/src/C3d.cpp
+++ b/c3d/src/C3d.cpp
@@ -24,6 +24,7 @@
 #include "clang/Sema/Lookup.h"
 #include "clang/Sema/ParsedAttr.h"
 #include "clang/Sema/Sema.h"
+#include "clang/Sema/Scope.h"
 #include "clang/Sema/SemaDiagnostic.h"
 #include "llvm/IR/Attributes.h"
 #include "llvm/Support/Debug.h"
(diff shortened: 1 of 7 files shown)


c3d/src/C3d.cpp
```diff
@@ -432,16 +433,127 @@ class C3dParameterAttrInfo : public C3dSimpleSpelling, C3dDiagOnStruct {
 
         State = SeenExpr;
         return AttributeApplied;
+     }
+
+     default:
```
msprotz, line 438: the C++ compiler should be able to prove exhaustivity of the case here... did you get a warning with your compiler?

c3d/src/C3d.cpp
```diff
@@ -432,16 +433,127 @@ class C3dParameterAttrInfo : public C3dSimpleSpelling, C3dDiagOnStruct {
...
+        Sema &S = P->getActions();
+
+        // A 'void' type
+        const Type *voidTy = S.getASTContext().VoidTy.getTypePtr();
+        QualType R = QualType(voidTy, 0);
+
+        // A dummy 'this' identifier
+        IdentifierInfo &IDD = P->getPreprocessor().getIdentifierTable().getOwn("this");
+
+        // Create a variable declaration for 'this', at 'void' type, and push it.
```
msprotz, line 495: is there any way to factor out this bit into a helper function? (not sure) it appears in a few places in the code

c3d/src/C3d.cpp
```diff
@@ -544,21 +659,43 @@ class C3dVisitor : public RecursiveASTVisitor<C3dVisitor> {
...
+    // since that's what 3d expects.
+    if (! WhereClauses.empty()) {
+        enum { First, Mid } State = First;
+        const int shift = strlen("c3d_where:");
+        Out << "\nwhere ";
+        for (const auto& W: WhereClauses) {
+            switch (State) {
+                case Mid:
+                    Out << " && ";
+                    // fall through!
```
msprotz, line 688: https://en.cppreference.com/w/cpp/language/attributes/fallthrough ?

c3d/src/C3d.cpp
```diff
@@ -544,21 +659,43 @@ class C3dVisitor : public RecursiveASTVisitor<C3dVisitor> {
...
+
+                case First:
+                    Out << "(" << W->getAnnotation().slice(shift, W->getAnnotation().size()) << ")";
```
msprotz, line 691: this adds superfluous parentheses when there's only a single where clause? is this desirable? (quite possibly)

c3d/src/Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: test
+
+ifeq (${CLANG_BUILD},)
+$(error "Please define $$CLANG_BUILD, it should point to your clang build directory")
+endif
+
+test:
+	make -C out/
+	${CLANG_BUILD}/bin/clang -mllvm -debug-only=c3d -std=c2x -fplugin=./out/C3d.so ../tests/basic0.h
```
msprotz, line 9: can you ifdef on the platform, e.g.

```
ifeq($(OS),Windows_NT)
SO = dll
else
  ifeq($(uname -s),Darwin)
    SO = dylib
  else
    SO = so
  endif
endif
```

or something along those lines?

c3d/src/C3d.cpp
mtzguido: I did actually, but now I see it was GCC:
```
/home/guido/r/everparse/c3d/src/C3d.cpp: In member function ‘virtual clang::ParsedAttrInfo::AttrHandling {anonymous}::C3dParameterAttrInfo::handleDeclAttribute(clang::Sema&, clang::Decl*, const clang::ParsedAttr&) const’:
/home/guido/r/everparse/c3d/src/C3d.cpp:442:3: warning: control reaches end of non-void function  -Wreturn-type]
  442 |   }
      |   ^
```
I suppose it's more defensive against badly-casted enum values?

mtzguido: Neat, thanks!

c3d/src/C3d.cpp
mtzguido: Yes, some bits of it at least, will do in future commits.

c3d/src/C3d.cpp
mtzguido: Neat, thanks!

c3d/src/C3d.cpp
msprotz: ah I'm compiling the plugin with clang, which is happy here -- any thoughts on whether we should require clang?

c3d/src/C3d.cpp
mtzguido: I wasn't super sure if it's OK to not wrap them. I think it is, since the thing inside must be a plain expression at this point, but I'd like to double check with the grammars before removing this.

c3d/src/Makefile
mtzguido: Sure, will do!

c3d/src/C3d.cpp
mtzguido: Let's call clang the canonical compiler? I'll switch to it here, but could also check now and then for GCC.

mtzguido:
> Regarding pushing scopes. I assume you would want to keep internal state to the plugin, relying on the invariant that the fields are visited in order, and knowing that upon entering everparse::process, you can "clear" the internal scope.
That makes sense. It does mean it will not be reentrant though, right? Just checking, don't think it's a problem for us.

> Is there any way you can keep the scope as a parser scope internally (rather than rolling out, say, your own vector) making sure it only grows, and pushing a scope that already contains the previous fields? That would require some sort of `enterNewPreFilledScope` API for the parser, which I'm not sure exists.
There is something similar for template scopes.. but not for normal ones, so we'd have to implement that. It doesn't seem terribly hard, but it's a change to clang. Maybe we should roll a vector for now, which should always work, and try this in the background?

> Is your scope management compatible with pretty-printing, i.e. does this still print a decent .h file?
I think it doesn't affect it at all, and the push/pop only happens for parsing constraints anyway.

msprotz:
> > Regarding pushing scopes. I assume you would want to keep internal state to the plugin, relying on the invariant that the fields are visited in order, and knowing that upon entering everparse::process, you can "clear" the internal scope.
> 
> That makes sense. It does mean it will not be reentrant though, right? Just checking, don't think it's a problem for us.
if there's a nested struct within a struct, we wouldn't be able to detect moving from the inner to the outer struct, e.g.

```
typedef struct {
  struct {
    int x, y, z;
  } origin;
  int angle;
} unary_vector;
```

we wouldn't be able to figure out we're moving from z to angle
> 
> > Is there any way you can keep the scope as a parser scope internally (rather than rolling out, say, your own vector) making sure it only grows, and pushing a scope that already contains the previous fields? That would require some sort of `enterNewPreFilledScope` API for the parser, which I'm not sure exists.
> 
> There is something similar for template scopes.. but not for normal ones, so we'd have to implement that. It doesn't seem terribly hard, but it's a change to clang. Maybe we should roll a vector for now, which should always work, and try this in the background?
yes, also I assume this kind of change would be even more specific and a hard sell for clang, so let's stick with a vector of names for now as a global that's shared between various plugins
> 
> > Is your scope management compatible with pretty-printing, i.e. does this still print a decent .h file?
> 
> I think it doesn't affect it at all, and the push/pop only happens for parsing constraints anyway.

good! you may want to call the `dump` method on the astcontext to check that clang is happy with the internal representation (should be ok -- I'm assuming the API is crafted in a way that prevents you from constructing inconsistent ASTs)

PR #36, opened 2020-11-06
Zero-terminated C strings

tahina-pro:
`T field_name[:string]` 
`T field_name[:string-byte-size-at-most size]` 

where `T` is the element type, and `size` is the size in bytes (in the latter case, the input buffer will consume `size` bytes but is expected to start with a maybe shorter zero-terminated string.)

The zero terminator is automatically deduced from `T` (currently assuming that `T` is a numerical type) because I claim that it would be too heavy to mandate the user to manually specify the zero every time they want a string.


nikswamy:
Thanks! THis is very useful. 

How about using the qualifier keywords "zero-terminated-string" and "zero-terminated-string-at-most-byte-size". It's long but very explicit

src/3d/Ast.fst
```diff
@@ -286,12 +286,18 @@ type array_qualifier =
   | ArrayByteSizeAtMost //[:byte-size-at-most
   | ArrayByteSizeSingleElementArray //[:byte-size-single-element-array
 
+noeq
+type field_array_t =
+  | FieldArrayNormal
```
nikswamy, line 291: this case indicates that the field is not an array, right?

tahina-pro: yes, the field is neither a string nor otherwise an array

tahina-pro:
> How about using the qualifier keywords "zero-terminated-string" and "zero-terminated-string-at-most-byte-size". It's long but very explicit

I'll do `zeroterm`. (I also proposed `cstring` but usually C strings are only byte strings.)

nikswamy:
thanks! merging

PR #40, opened 2021-01-22
3d support for big-endian unsigned integers

tahina-pro:
Both big-endian and little-endian unsigned integers are already supported in LowParse and QuackyDucky, but only little-endian unsigned integers are supported in 3d currently. This PR extends support for big-endian unsigned integers to 3d.

In 3d, `UINT16`, `UINT32` and `UINT64` are little-endian unsigned integers, and now, thanks to this PR, `UINT16BE`, `UINT32BE`, `UINT64BE` are big-endian unsigned integers.

To add such support, it was necessary to extend 3d: while I added such types in the surface syntax, I didn't need to add new integer types in the AST. Indeed, from my understanding, for leaf types, the names of the parsers, validators and readers from `src/3d/prelude` are inferred from the surface syntax rather than the AST. @nikswamy Please comment and correct me if I'm wrong.

Then, reusing the same AST integer types for big-endian and little-endian integers allows me to seamlessly cast between one another.

I accordingly added a test in `src/3d/tests/Arith.3d`.


tahina-pro:
I don't know why 70c37c67fb5e4a7131322f29c5419f6d1e78401b is not appearing here, it should

tahina-pro:
> I don't know why [70c37c6](https://github.com/project-everest/everparse/commit/70c37c67fb5e4a7131322f29c5419f6d1e78401b) is not appearing here, it should

Finally that commit did appear, but only after I pushed an empty commit.

PR #42, opened 2021-02-05
Nik check dep exists

nikswamy:
Given 

```
typedef A.t At
```
where A is unbound, this now fails with

```
.\Constants.3d:(15,9): (Error) Dependency not found: A
```

### DIFF
--- a/src/3d/Deps.fst
+++ b/src/3d/Deps.fst
@@ -13,6 +13,9 @@ type dep_graph = list edge
 let all_edges_from (g:dep_graph) (node:string) : Tot (list edge) =
   List.Tot.filter (fun (src, _dst) -> src = node) g
 
+let dep_exists dirname name =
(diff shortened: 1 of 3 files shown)


tahina-pro (approved):

aseemr:
Thanks Nik! I have merged the PR. We can do some minor improvements, e.g.

```
aseemr@ASEEMR-SBOOK ~/everest/quackyducky/src/3d/tests
$ cat Test.3d
module A = TestA

entrypoint
typedef struct _T {
  A.MT  f;
} T;

aseemr@ASEEMR-SBOOK ~/everest/quackyducky/src/3d/tests
$ ../3d Test.3d
.\Test.3d:(5,3): (Error) Dependency not found: TestA
```

Instead of giving the error on Line 5, in terms of the full module name, we could give the error when the module abbreviation is defined.

I will do this small tweak later. Thanks again!

Issue #16, opened 2020-06-14
3d: remove --module_name


tahina-pro:
Suggested by @nikswamy at https://github.com/project-everest/everparse/pull/15#discussion_r436972475

PR #45, opened 2021-02-11
Generate a Makefile for 3d parallel processing

tahina-pro:
## Motivation

So far `3d --batch` has been marred with slow execution time due to processing everything sequentially. To avoid that, this PR offers the user the option to produce a Makefile instead, so that the following process:

1. The user invokes `3d --makefile gmake --makefile_name MyMakefile Source1.3d Source2.3d`
2. The user invokes `make -j -f MyMakefile`

should be equivalent to `3d --batch Source1.3d Source2.3d` but with parallelism.

## New `3d` options

This new feature relies on new 3d options that I just added as part of this PR:

* `--makefile gmake`: generate a GNU Makefile. If this option is set, then nothing else than that Makefile is generated (not even the .fst, Wrapper.c, etc. files). The command generated for each Makefile rule uses `$(EVERPARSE_CMD)` as a placeholder for the `3d`/`everparse.cmd`/`everparse.sh` invocation, and `$(EVERPARSE_OUTPUT_DIR)` for the output directory.
* `--makefile_name MyMakefile` : set the full path and name of such generated Makefile to MyMakefile. Default is `EverParse.Makefile` in the output directory specified by `--odir`
* `--__micro_step verify` : proceed to the verification of a `.fst` or `.fsti` file, producing a `.checked`  file. The options passed to F* are the same as those that would have been passed with `3d --batch`; this allows us to avoid duplicating them in the code of 3d or in the produced Makefile. Then, 3d expects a list of `.fst` or `.fsti` files as arguments instead of `.3d` files.
* `--__micro_step extract` : proceed to the extraction of a `.fst`  file, producing a `.krml`  file, similarly.
* `--__produce_c_from_existing_krml`: proceed to the extraction of the `.c` and `.h` files. Then, 3d expects a list of `.3d` files and will perform ~one single call to KReMLin~ one call to KReMLin per such .3d file. Basically, this option performs everything starting from the krml invocation, including calls to clang-format and copyright headers.
* `--__skip_deps` : assume that the .3d files provided as arguments to `3d` are all given in the right dependency order, thus skipping any dependency analysis. This is ignored if `--makefile` is provided, because in that case I need to compute a dependency graph anyway.

## Code changes

In terms of code changes, this pull request does all of the following:
* A new module, `GenMakefile`, to generate the Makefiles.
* Slight refactor of batch operations: I isolated each operation needed for the `--__micro_step` options above
* `--print_in_place` is now always performed, independently of `--batch`
* Slight refactor of `Dep`: I expose some operations on the dependency graph
* Deep refactor of `Options` : 
  - I now classify 3d options into categories
  - for each option, say `--foo`, I automatically create its opposite option, `--no_foo`, to untoggle a boolean option, reset a string option or empty a list option.
  - I clarify the dependencies between options, in the sense that some options imply others (for instance, if the user sets `--clang_format_executable` , then `--clang_format` and `--batch`  are automatically set.)
  - all options starting with `__` are hidden from the `--help` message
  - the `--help` message now prints the status of options as a normalized sequence of options, so that calling 3d with this sequence should have the same behavior.
* Add the options described above.
* Changed the behavior of the src/3d/tests/modules test to generate a Makefile and use it instead of `--batch`

## Next steps

It now remains to:

* [x] Replace the directory names of the source files with a variable, say `EVERPARSE_SOURCE_DIR`.
* ~Maybe find a way to output the options given to `3d --gnu_makefile` into the Makefile, to ensure that all calls to F* are consistent with those 3d options given by the user. (Basically, this is one of the main reasons why I refactored `Options`: to be able to print such list of options.)~ Won't do; the generated Makefile has rules call EverParse through the EVERPARSE_CMD variable, which I claim is enough for the user to specify their own options. With that variable, EverParse options can be enforced to be the same as those used to generate the Makefile itself, see https://github.com/project-everest/everparse/blob/6590036282868aa915f0dbcee293ccf002db7eb2/src/3d/tests/modules/Makefile
* [x] change the structure of the `--gnu_makefile` option, becoming `--makefile gmake|nmake`
* [x] Add a backend to produce NMakefiles
* ~Maybe add a backend to produce Ninja files~ Postponed, cf. #48 
* [x] Probably rename the unrelated `--skip_makefiles` option to something like `--skip_c_makefiles` (this option controls whether 3d should remove the Makefiles produced by krml to _compile_ the generated .c files)
* [x] ~Maybe~ figure out how to replace the single KReMLin call with one invocation of KReMLin per `.3d` file (true separate compilation) ~(hard)~ (finally it was quite easy, thanks @msprotz !)
* [x] Figure out how to replace the single call to 3d to generate .fst/.fsti/*Wrapper files with one call to 3d per `.3d` file (true separate compilation.) ~I may need some help from @nikswamy  here.~ Thank you @aseemr !
* [x] Take `StaticAssertions.c` files into account when they exist. Currently, the produced Makefile produces them but ignores them.
* [x] Support non-existing `Wrapper.h` and `Wrapper.c` files (for modules with no `entrypoint`)
* [x] Add the copyright file to the dependencies of *.h and *.c files, if it exists


tahina-pro:
By the way, this PR, via 38538d9, would also solve #16.

tahina-pro:
Whenever 3d needs to run clang-format, it first copies a local .clang-format to the generated directory. Then, with the Makefile, such copy step is replicated for each generated .c/.h file, which introduces a race. As far as I know, there is no way to tell clang-format to use a style file different from .clang-format (cf. https://bugs.llvm.org/show_bug.cgi?id=20753). So, I need to:
* [ ] Introduce a micro-step to copy .clang-format and add a Makefile rule to run it

tahina-pro:
As of now, Makefiles generated by `--makefile gmake`  happen to also be compatible with NMAKE by virtue of not using anything like implicit variables or any features specific to GMake. Nevertheless, I introduced `--makefile nmake` to generate NMAKE-compatible Makefiles, as a future-proof option, although it currently does the exact same thing as `--makefile gmake`.

PR #53, opened 2021-04-24
3d support for non-strong prefix parsers, all_bytes, all_zeros

tahina-pro:
This PR adds 3d support for parsers that do not necessarily have the strong prefix property.
* The 3d "typechecker" (along with prelude combinator types) enforces that the first part of a (dependent or non-dependent) pair have the strong prefix property.
* New `all_bytes` type and parser consumes all of the remaining bytes of the current input buffer, always succeeds.
* New `all_zeros` type and parser consumes all of the remaining bytes of the current input buffer, fails if one of those bytes is nonzero.
* Example in `FieldDependence0.3d`: list of tagged unions where the payload of one of those cases is `all_zeros`. That case models the terminator symbol, so that for a given buffer length, the list shall contain either only elements that together fit that size, or elements followed by the terminator symbol and padding of zeros.

Thus, this PR supersedes and generalizes branch `taramana_3d_dep_pair_with_terminator` where that example was defined as an ad-hoc combinator in 3d prelude. With this PR, such a datatype can now be defined in a 3d file.


nikswamy:
Thanks @tahina-pro ... this is really great to have. I successfully used it in the TCP header spec to encode the options terminated by the END_OF_LIST marker. It was really smooth.

PR #55, opened 2021-06-24
Fine-grained, user-controllable error reporting

nikswamy:
# The problem

Consider the following example

```
typedef struct _T
{
   UINT32 x
   {
       x >= 17
   };
   UINT32 y
   {
       y >= x
   };
} T;
```

When defining a validator for this type, we generate:

1. field codes: Every field in every type of a 3d file has a unique
   code, e.g.,  `T.x = 1` and `T.x = 2`.

2. Error codes: EverParse has a fixed set of error codes,
   corresponding to its different failure modes, including codes for:

```
    1, "generic error";
    2, "not enough data";
    3, "impossible";
    4, "list size not multiple of element size";
    5, "action failed";
    6, "constraint failed";
    7, "unexpected padding";
    8+, "unspecified";
```

If we fail when validating, say, `T.x`, we encode in the return value
from EverParse three things:

 1. The position in the input buffer where the failure occurred
 2. The code of the field on which the error occurred.
 3. The error code

At the top-level, we decode this information and pass it to a
top-level user-provided error handler, whose signature is

```
void
EverParseError(char *structName,
               char *fieldName,
               char *errorReason);
```

Some limitation of this scheme is that it is not context sensitive

We call back a single error handler at the top-level. It is hard to
fully reconstruct the failure from this error. For example, consider
this:

```
typedef struct _S {
  T t1;
  T t2;
} S;
```

A failure in `S.t1.x` cannot be distinguished from `S.t2.x`.

Some additional drawbacks:

 * The position in the input buffer at which the error occurred is not
   passed to the callback.

# The solution: on-error callbacks

Rather than encoding the error and field in the result code and
propagating it to the top-level callback, we will instead allow the
user to specify an error callback to be invoked repeatedly at every
level of the validation stack.

There are a few elements to the solution:

1. The type actions and validators are revised to take an additional `app_ctxt` parameter, representing some application chosen context type (e.g., some void* that the context passes in). The revised type of actions is:

```
let action p inv l on_success a
  = ctxt:app_ctxt -> //<---new
    #len: U32.t ->
    sl: input_buffer_t len ->
    pos:LPL.pos_t ->
    res:U64.t ->
    Stack a
      (requires fun h ->
        let open LPL in
        LPL.live_input_buffer h sl /\
        B.live h ctxt  /\ <-- context is initially live
        inv (LPL.slice_of sl).LPL.base h /\
        loc_not_unused_in h `loc_includes` (app_loc ctxt l) /\
        address_liveness_insensitive_locs `loc_includes` (app_loc ctxt l) /\ //<--- it cannot be freed by us
        (app_loc ctxt l) `loc_disjoint`  //<-- it's disjoint from the validation buffers
        (loc_buffer (LPL.slice_of sl).base `loc_union` R.loc_perm (LPL.perm_of sl)) /\
        (U64.v res <= U64.v validator_max_length ==>
           valid_pos p h (LPL.slice_of sl) (uint64_to_uint32 pos) (uint64_to_uint32 res)) /\
        (on_success ==> U64.v res <= U64.v validator_max_length))
      (ensures fun h0 _ h1 ->
        modifies (app_loc ctxt l) h0 h1 /\ //we may modify it 
        B.live h1 ctxt /\         //it remains live
        h1 `extends` h0 /\
        inv (LPL.slice_of sl).LPL.base h1)
```

2. Every 3D validator is now additionally parameterized by an error handler of type:

```
let error_handler = typename:string ->
                    fieldname:string ->
                    error_reason:string ->
                    action (parse_impos()) true_inv eloc_none false unit
```
and a ``app_ctxt``

3. Every validator is wrapped with a call to validate_with_error_handler, that calls the user-provided call back in case a validator fails, passing in the typename, fieldname, error reason, context, position, etc.

4. At the top-level, the user can instantiate the context parameter as
   they wish and link it with their implementation of the extern. 

By instantiating error callback extern appropriately, the user can
effectively reconstruct the entire validation call stack when an error
occurs, or a stack up to a bound.

For example, the caller can

1. Stack allocate a context up to some bound

```
typedef struct _ErrorFrame
{
    uint32 position;
    char *typename;
    char *fieldname;
    error_reason_enum_t code;
} ErrorFrame;

typedef struct _Frames {
    UINT8 ctr;
    UINT8 nframes;
    ErrorFrame frames[0];
} Frames;

void OnErrorCallback (Frames *frames,
                      uint32 position,
                      const char *fieldname,
                      const char *typename,
                      error_reason_enum_t code)
{
   if (frames->ctr < frames->nframes)
   {
       frames->frames[frames->ctr] =
        ((ErrorFrame)
          {
             .position = position;
             .typename = typename;
             ...
          });
       frames->ctr++;
   }
}
```

And to call into a validator instantiating the context like so:

```
{
   ErrorFrame frames[10];
   Frames ctxt =
    {
       ctr = 0;
       nframes = 10;
       .frames = frames;
    };

   validateT(&ctxt, ...)
}

```

# Top-level wrapper

To maintain backwards compat, and to provide a default error handler, the top-level wrapper provides something like this, which only tracks a single error frame and can report the error using the same user-provided error reporting function as before.

```
void ErrorHandlingEverParseError(char *StructName, char *FieldName, char *Reason);

typedef struct _ErrorFrame
{
	BOOLEAN filled;
	uint32_t start_pos;
	uint32_t end_pos;
	char *typename;
	char *fieldname;
	char *reason;
} ErrorFrame;

void DefaultErrorHandler(
	EverParseString typename,
	EverParseString fieldname,
	EverParseString reason,
	uint8_t *context,
	uint32_t len,
	uint8_t *base,
	uint64_t start_pos,
	uint64_t end_pos)
{
	ErrorFrame *frame = (ErrorFrame*)context;
	if (!frame->filled)
	{
		frame->filled = TRUE;
		frame->start_pos = start_pos;
		frame->end_pos = end_pos;
		frame->typename = typename;
		frame->fieldname = fieldname;
		frame->reason = reason;
	}
}

BOOLEAN ErrorHandlingCheckS(uint8_t *base, uint32_t len) {
	ErrorFrame frame;
	frame.filled = FALSE;
	uint64_t result = ErrorHandlingValidateS((uint8_t*)&frame,  &DefaultErrorHandler, len, base, 0);
	if (EverParseResultIsError(result))
	{
		if (frame.filled)
		{
			ErrorHandlingEverParseError(frame.typename, frame.fieldname, frame.reason);
		}
		return FALSE;
	}
	return TRUE;
}
```

# Some final remarks

One might be concerned about performance.

In case of ill-formed packets, for deep packet formats, we may call
the callback many times as we pop the validation stack.

This may impose some overhead, but the cost of it, only on the error
path, should be bounded by the static nesting depth of the format
specification.

Besides, even in the case of denial-of-service attacks from attackers
sending many malformed packets, the cost of rejecting many packets
even with elaborated error reporting will be usually far negligible
compared to the total processing cost (post-validation and
post-actions) of one valid packet.

At the very worst, in case this is a concern, we could even limit such
advanced reporting to debug builds only.

### DIFF
--- a/doc/3d-snapshot/BF.c
+++ b/doc/3d-snapshot/BF.c
@@ -2,7 +2,24 @@
 
 #include "BF.h"
 
-static inline uint64_t ValidateDummyEmp(uint64_t StartPosition)
+static inline uint64_t
+ValidateDummyEmp(
+  uint8_t *Ctxt,
+  void
+  (*Err)(
+    EverParseString x0,
+    EverParseString x1,
+    EverParseString x2,
+    uint8_t *x3,
+    uint32_t x4,
+    uint8_t *x5,
+    uint64_t x6,
+    uint64_t x7
+  ),
+  uint32_t Uu,
+  uint8_t *Input,
+  uint64_t StartPosition
+)
 /*++
     Internal helper function:
         Validator for field _dummy_emp
@@ -10,13 +27,56 @@ static inline uint64_t ValidateDummyEmp(uint64_t StartPosition)
 --*/
 {
   /* Validating field emp */
-  return StartPosition;
+  uint64_t positionAfterDummy = StartPosition;
+  if (EverParseIsSuccess(positionAfterDummy))
+  {
+    return positionAfterDummy;
+  }
+  Err("_dummy",
+    "_dummy_emp",
+    EverParseErrorReasonOfResult(positionAfterDummy),
+    Ctxt,
+    Uu,
+    Input,
+    StartPosition,
+    positionAfterDummy);
+  return positionAfterDummy;
 }
 
-uint64_t BfValidateDummy(uint32_t InputLength, uint8_t *Input, uint64_t StartPosition)
+uint64_t
+BfValidateDummy(
+  uint8_t *Ctxt,
+  void
+  (*Err)(
+    EverParseString x0,
+    EverParseString x1,
+    EverParseString x2,
+    uint8_t *x3,
+    uint32_t x4,
+    uint8_t *x5,
+    uint64_t x6,
+    uint64_t x7
+  ),
+  uint32_t Uu,
+  uint8_t *Input,
+  uint64_t StartPosition
+)
 {
   /* Field _dummy_emp */
-  return ValidateDummyEmp(StartPosition);
+  uint64_t positionAfterDummy = ValidateDummyEmp(Ctxt, Err, Uu, Input, StartPosition);
+  if (EverParseIsSuccess(positionAfterDummy))
+  {
+    return positionAfterDummy;
+  }
+  Err("_dummy",
+    "emp",
+    EverParseErrorReasonOfResult(positionAfterDummy),
+    Ctxt,
+    Uu,
+    Input,
+    StartPosition,
+    positionAfterDummy);
+  return positionAfterDummy;
 }
 
 void BfReadDummy(uint32_t InputLength, uint8_t *Input, uint32_t StartPosition)
--- a/doc/3d-lang.rst
+++ b/doc/3d-lang.rst
@@ -640,6 +640,89 @@ using a fully qualified name of the form ``<MODULE NAME>.<IDENTIFIER>``.
 A commented example is available `in the EverParse repository
 <https://github.com/project-everest/everparse/blob/master/src/3d/tests/modules/>`_.
 
+Error handling
+--------------
+
+When a validator fails, EverParse supports invoking a user-provided
+callback with contextual information about the failure.
+
+An error handling callback is a C procedure with the following signature:
+
+.. code-block:: c
+
+  typedef void (*ErrorHandler)(
+    const char *TypeName,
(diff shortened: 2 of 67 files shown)


nikswamy:
I fixed my description to resolve the open issue of an error handler per module. Instead, error handler are provided dynamically and threaded down through the validation stack, making it usable in a multi-module setting.

I've also update 3d-lang.rst with documentation of error handling.

doc/3d-snapshot/BF.c
```diff
@@ -2,21 +2,81 @@
...
+)
 /*++
     Internal helper function:
         Validator for field _dummy_emp
         of type BF._dummy
 --*/
 {
   /* Validating field emp */
-  return StartPosition;
+  uint64_t positionAfterDummy = StartPosition;
```
nikswamy, line 30: @aseemr @tahina-pro : Here's an always-succeeding unit validator that ended up getting wrapped with an error handler

PR #58, opened 2021-07-02
Validate from non-contiguous input "streams"

tahina-pro:
# The problem

So far all applications of EverParse have been reading and validating input bytes from a single contiguous buffer.

However, other applications present input data using different, maybe non-contiguous, data structures. So, to handle them, we aim to evolve EverParse so that it too is parametric in a user-chosen input layout.

We want to handle such data structures without integrating their specifics into the actual implementations of EverParse validators.

# This PR: input "stream" interface

We propose to design an interface to the input buffer as a "stream". The only operations we need are:

- `bool has(input_buffer_t buf, uint32_t n);`

  Returns `true` if and only if the input buffer `buf` has at least `n` bytes available, but without reading them or advancing the input buffer.

- `void read(input_buffer_t buf, uint32_t n, uint8_t *out);`

  reads `n` bytes into `out` and advances the input buffer so that the next call to that function will read the next byte (thus ensuring absence of double-fetches.)

- `void skip(input_buffer_t buf, uint32 count);`

  advances the input buffer by `count` bytes, without reading them. Useful when validating an integer field when its actual value does not matter.

# Status as of 2021-07-01

At this point, I provide a Low* interface for input streams and a verified implementation for the "flat buffer" case, and I have reimplemented validators and actions (except `field_ptr`) following this interface. The next steps are:

* [x] Recover `field_ptr`, but only in the case of the "flat buffer" implementation
* [x] Allow linking with an `extern` input stream implementation, and allow the user to choose between that and the verified "flat buffer" implementation
* [x] Widen some integers: `get_read_count`, arguments to `has`, `read`, `skip`, etc.
* [x] Avoid double copies for readers (one coming from `read`, another coming from kremlib (https://github.com/FStarLang/kremlin/blob/986bd298eecedc96278e1ba712fadd7c5f7d276b/include/kremlin/lowstar_endianness.h#L157 sqq.)


tahina-pro:
07890a4974aa8659b1eaf7b6e480ed5feb167d75 is not very satisfactory, since it requires a lot of `friend`ing, and also removes the need for `read` in the type class. So I'm going to force-push with a more generic version of `read` that either returns a slice of the input buffer or copies its contents to the destination buffer provided by the user.

tahina-pro:
Compared to `master`, the error handler lost track of the end position, i.e. the number of bytes read from the input buffer at the time the error is raised. In fact, it turns out that such number of bytes read can be tracked by validators just as before, instead of a reference attached to the input stream. So I should:
* [ ] Remove get_read_count and have the position tracked by validators

tahina-pro:
I can now link with a custom unverified input stream implementation, but such a custom implementation has to provide a `struct EverParseInputStream_s *` type, which is quite restrictive. Instead, I should:
* [ ] Ask for a header file name (via EverParse command-line options) to `#include` into every produced .h file; this header file is supposed to contain a `typedef ... EverParseInputStream` declaration.

PR #41, opened 2021-01-23
Modules in 3d

aseemr:
See `tests/modules` for a testcase.

PR #62, opened 2021-10-29
A new interpreter and partial-evaluation-based approach for EverParse 3D

nikswamy:
This is a draft PR, not yet ready for merging. Meant more to collect a worklist of items that must be completed before merging. 

## Essential items

- [x] Complete and merge the F* branch nik_erasable into F*. It supports erasing erasable arguments in the normalizer and nbe (nik)
      Thanks to Tahina for fixing Benton2004, LowParse.Writers, etc.

- [x] Support for Enum typed parameters, by adding, at least within the internal syntax, a first-class notion of refinement type (at the moment, refinements are only allowed on the fields of a struct, not in parameter types). This too is not fully supported on master, but it should be. (nik)

- [x] Remove the interpret_alt flag and retain --interpret only. (nik)
 
- [x] Adding back support for error handling wrappers on generated validators (nik)

- [x] Fix a bug where type aliases to defined struct or case types don't work (nik)

- [x] Add an interface file for interpreter mode so that Kremlin only generates a .h for the specified entrypoint, rather than for all functions (nik)

- [x] Propagating variable names from the source for rename_let: currently they are all harcoded to "a". (nik)
       -- Support for this has been enabled and it is propagated to the C code
       -- However, I notice in some cases that the field name is still missing when emitted by the 3d source toolchain

- [x] propagate comment strings for refinements
 
- [x] Restoring batch mode (tahina)

- [x] Restoring makefile mode to support compiling the generated C files (tahina)

- [x] Support for Action_call in Interpreter.fst (nik)

- [x] Support for output types based on it Action_call (nik)

- [x] Make Interpreter generic with respect to buffer vs extern (Tahina)
   - In EverParse3d.Actions.All, for backwards compat, can we have an action_field_pos returning a U32.t, for the buffer case
   - Rename existing action_field_pos to action_field_pos64

- [x] Remove a spurious stack allocation in one of the buffer combinators (Tahina)

## Other improvements

- [ ] Fixing F* --dep so that it does not emit .krml targets for files that are explicitly excluded from extraction using the --extract option (nik)
   - We have an F* issue and branch with a fix. Trying to get an everest green
   - Meanwhile, we have worked around this behavior with a makefile hack
   
- [x] Support for dependence on Enum-typed fields (by inlining them). This is not supported on master anyway, but it should be. (nik)

- [x] In fact, in F* we should forbid using the empty string "" with rename_let. This led to strange failures in Kremlin. (nik)

- [ ] Revise the syntax of actions so that they don't have to be so explicitly monadic. 

- [ ] Separate :on-success actions from :check actions. The former return unit; the latter return bool and can cause validation to fail. This will save us from having to write `return true` everywhere in actions.


msprotz:
> Fixing F* --dep so that it does not emit .krml targets for files that are explicitly excluded from extraction using the --extract option (aseem)

This will break a large number of applications (e.g. Noise*, HACL*, and soon MLS*) that rely on this behavior.

In those applications, we do `--extract Spec.* -Impl.*`, meaning "extract specs to OCaml and ditch low-level code" -- this allows us to then test the specs via OCaml compilation. Low-level compilation is controlled by kremlin's bundle flags.

Note that changing this behavior might be dangerous, too, since for Low* code you may inadvertently remove a definition that would be needed by another file, resulting in `Warning 2` errors and potentially invalid C code if, for instance, `B.foo` needs `A.bar` but you forgot that and passed `--extract -A`.

Hope this helps,

Jonathan

tahina-pro:
Thank you Jonathan for sharing your concerns on skipping `.krml` files for F* `--dep`, let's continue that discussion in FStarLang/FStar#2385.

src/3d/Main.fst
```diff
@@ -201,15 +256,25 @@ let process_file (en:env) (fn:string) (modul:string) (emit_fstar:bool) (emit_out
...
 
   let ds = Binding.get_exported_decls en.binding_env modul in
   
   { binding_env = Binding.finish_module en.binding_env modul ds;
     typesizes_env = TypeSizes.finish_module en.typesizes_env modul ds;
-    translate_env = Translate.finish_module en.translate_env modul ds }
+    translate_env = 
+      if Options.get_interpret()
+      then en.translate_env
+      else Inl (Translate.finish_module (left en.translate_env) modul ds) }
```
tahina-pro, line 277: This seems to make `--interpret` break with modules (#41) in non-batch mode. More precisely, in `src/3d/tests/modules/src`, running `3d.exe Test.3d`, `3d.exe --batch Test.3d` or `3d.exe --batch --interpret Test.3d` all succeed (and in the non-batch case only produces .fst files for Test, as expected, by loading its .3d dependencies without producing F* code for those), but running `3d.exe --interpret Test.3d` fails with:
```
Fatal error: exception (Failure "Type decl not found for Point._TWO_D_POINT")
```
Without `--interpret`, when a single X.3d file is passed to 3d, its dependencies are loaded and their contents are added to the environment (in the `else` branch) but .fst files are generated only for X.
With `--interpret`, the environment does not seem to be changed, or is changed only if `.fst` files are actually generated, which is not the case for dependencies. Is my understanding correct? If so, then why do we need to strictly tie populating the environment with generating F* code?

tahina-pro: I think this comes from that line in `InterpreterTarget.translate_decls`:
https://github.com/project-everest/everparse/blob/617f21ebf635c83da457619d1225a110a21f5ef1/src/3d/InterpreterTarget.fst#L480
Is there any way to disentangle the call to `translate_decls` from `Main.emit_fstar_code_from_interpreter`?

nikswamy: thanks Tahina. I fixed this

PR #86, opened 2022-10-05
Add documentation and example for quackyducky part of Everparse

samuelchassot:
As I had to figure out how to use `quackyducky` during my research, I thought I could contribute to the documentation with what was missing.

I would be happy to get feedback :)


tahina-pro:
Thanks a lot Samuel for your proposal! Sorry for the late reply, our focus has long been on the 3d side, as you guessed right.
So far we have not made any distinctions between tests and examples, in the sense that I meant `tests/sample` as an (undocumented) example for QuackyDucky. Worse, 3d tests and examples are all in a separate directory, `src/3d/tests`.
So, restructuring all tests and examples is definitely on my radar. To this end, I will probably integrate your proposal as `examples/qd` and move some of `src/3d/tests/tcpip` into something like `examples/3d`.

PR #87, opened 2022-10-11
Fix for a change to term_eq in F*

mtzguido:
When FStarLang/FStar#2730 is merged, `term_eq` will have some slight differences in behavior. In particular it will not ignore universes any more so I believe most calls here will fail. Since these calls are meant to check if a term is equal to some particular top-level name, replace them with `is_fvar`.

**Note**: this function does not currently exist in F* (it is added by that same PR), so this should only be merged after that.

I got a local everest green with the updated F* and the patched everparse (no changes needed to other projects).

------

LowParse: tactics: remove calls to term_eq, use is_fvar instead

`term_eq` received some fixes in F* (such as honoring universes) that cause some failures here. The old behavior is accessible via a (Tac) primitive `term_eq_old`. but these calls are meant to decide if a term is equal to some given top-level name, so just use `is_fvar` instead.

### DIFF
--- a/doc/3d-snapshot/Color.h
+++ b/doc/3d-snapshot/Color.h
@@ -12,17 +12,17 @@ extern "C" {
 
 
 #include "EverParse.h"
-/*
+/**
 Enum constant
 */
 #define COLOR_RED ((uint32_t)1U)
 
-/*
+/**
 Enum constant
 */
 #define COLOR_GREEN ((uint32_t)2U)
 
-/*
+/**
 Enum constant
 */
 #define COLOR_BLUE ((uint32_t)42U)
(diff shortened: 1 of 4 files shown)


mtzguido:
Slight update, this now works with F* commit [20e3eede1fc3aeb4b5828c3661d4991161d2b03d](https://github.com/FStarLang/FStar/commit/20e3eede1fc3aeb4b5828c3661d4991161d2b03d) ("examples: native_tactics: fix").

mtzguido:
The changes to F* have made it to master, so this will be needed to fix the build. Heads up @tahina-pro

tahina-pro:
I have an Everest green. Thanks a lot Guido!

PR #88, opened 2022-11-11
Some changes related to the F* core typechecker

aseemr:
Related to https://github.com/FStarLang/FStar/pull/2747.


src/lowparse/LowParse.Spec.Tac.Sum.fst
```diff
@@ -111,13 +111,18 @@ let rec maybe_enum_destr_t'_tac () : T.Tac unit =
...
+                        then T.fail "Done"
+                        else (T.compute (); T.smt ())));
       T.qed ()
     end
     else T.fail "Unknown shape for l2"
   | _ -> T.fail "Not the rigt arguments to maybe_enum_destr_t'"
 
 noextract
 let maybe_enum_destr_t_tac () : T.Tac unit =
+  T.set_guard_policy T.Goal;
```
aseemr, line 125: The tactic now generates some smt goals because of stricter typechecking of the unification variables. The goals are all straightforward to solve by normalization. So the fix here is to set the policy as Goal, asking the tactic engine to present these guards as goals, and then adding a loop to norm those goals and discharge.

src/lowparse/LowParse.Spec.Tac.Sum.fst
```diff
@@ -111,13 +111,18 @@ let rec maybe_enum_destr_t'_tac () : T.Tac unit =
     if tl2_fun `T.is_fvar` (`%Nil)
     then begin
       T.apply (`maybe_enum_destr_nil);
+      ignore (T.repeat (fun _ ->
```
aseemr, line 114: This is the loop that norms typechecking related smt goals.

PR #93, opened 2023-03-14
Remove effect-based writers

tahina-pro:
In support of FStarLang/FStar#2851

F* indexed effects with informative arguments will no longer extract, so this PR removes them from LowParse, QuackyDucky, and their tests.


tahina-pro:
I have an Everest green: project-everest/everest@0b71dd25c445f61c6f50744db4dcc1a3bb8b8acc

PR #96, opened 2023-04-18
Update usage of --split_queries

mtzguido:
This is a fix needed to merge https://github.com/FStarLang/FStar/pull/2776. Will update later with an everest green including both changes.


mtzguido:
Everest green here https://github.com/project-everest/everest/actions/runs/4727379343

PR #106, opened 2023-10-25
Generate test checker executable

### LINKED ISSUE #105: Extend 3d test infrastructure to accept a module and a witness
The current test harness supports generating positive and negative test cases for a given parser using z3. 
However, it would be great if we could also provide a candidate test witness and a 3d file, to determine if the generated parser would accept or reject the provided test witness. 

Here is an example of a DHCP packet in hex format that can be passed into the harness as a .pcap file (or other formats):

``` 01010600a27af44c0000000000000000000000000000000000000000001cc0e8232100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000638253633501017401013d0701001cc0e823213204ac1014680c094b464943542d4c33313c084d53465420352e30370b010f03062c2e2f1f21f92b2b02dc00ff ```

### PR DESCRIPTION
**Pull request by tahina-pro**
This PR introduces 3D option `--test_checker`, which, when followed by a parser name of the form `Modulename.parsername`, generates a test executable, `test.exe`, so that `./test.exe filename.dat arg1 arg2...` opens the file `filename.dat` and runs the validator for `Modulename.parsername`, with arguments `arg1`, `arg2`..., on the binary data contained in that file, and returns 0 if the data in the file passes validation, 1 if it fails validation, 2 for any other error (wrong number of arguments, file is missing, cannot mmap, etc.)

The file passed to `test.exe` is assumed to contain all of the input data to be passed to the validator, in binary form. If the user wants to start from some `filename.hex` that contains hex data instead (e.g. .pcap), it is the responsibility of the user to first run something like `xxd -r -p filename.hex filename.dat ` to convert that hex file to a binary file before passing the latter to `test.exe`

This should fix #105

### DIFF
--- a/doc/3d-snapshot/BF.c
+++ b/doc/3d-snapshot/BF.c
@@ -57,53 +57,55 @@ ValidateBf2bis(
   uint16_t bitfield0 = (uint16_t)(uint32_t)r0;
   /* Checking that we have enough space for a UINT16, i.e., 2 bytes */
   BOOLEAN hasBytes1 = 2ULL <= (InputLength - positionAfterBitfield0);
-  uint64_t positionAfternone;
+  uint64_t positionAfterBitfield1;
   if (hasBytes1)
   {
-    positionAfternone = positionAfterBitfield0 + 2ULL;
+    positionAfterBitfield1 = positionAfterBitfield0 + 2ULL;
   }
   else
   {
-    positionAfternone =
+    positionAfterBitfield1 =
       EverParseSetValidatorErrorPos(EVERPARSE_VALIDATOR_ERROR_NOT_ENOUGH_DATA,
         positionAfterBitfield0);
   }
   uint64_t positionAfterBf2bis0;
-  if (EverParseIsError(positionAfternone))
+  if (EverParseIsError(positionAfterBitfield1))
   {
-    positionAfterBf2bis0 = positionAfternone;
+    positionAfterBf2bis0 = positionAfterBitfield1;
   }
   else
   {
     uint16_t r = Load16Le(Input + (uint32_t)positionAfterBitfield0);
-    uint16_t none = (uint16_t)(uint32_t)r;
+    uint16_t bitfield1 = (uint16_t)(uint32_t)r;
     BOOLEAN
-    noneConstraintIsOk =
-      EverParseGetBitfield16(none,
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by nikswamy** [comment]
Thanks for adding the error handler!

**Comment by tahina-pro** [comment]
The generated test checker now rejects test files that contain data beyond a valid object: in other words, if the validator succeeds but does not read all input, then the test checker now considers the test failing, and returns status code 1. Validator failure returns status code 2, and 3 for any other kind of errors.

**Comment by tahina-pro** [comment]
The `--z3_test` and `--z3_diff_test` modes now output each test case generated by Z3 into a raw binary file, whose name is of the form `witness.nnn.POS.validatorname.arg1.arg2.dat` and `witness.nnn.NEG.validatorname.arg1.arg2.dat` for positive and negative test cases generated by `--z3_test` , and `witness.nnn.POS.validatorname1.NEG.validatorname2.arg1.arg2.arg3.dat` for differential test cases generated by `--z3_diff_test`. Those are raw binary files, so it is the responsibility of the user to run `xxd -p` on those files to convert them to hexadecimal format.

**Comment by lemmy** [comment]
There seems to be a off-by-one of some sorts s.t. the first negative witness is actually a positive one.

```bash
@lemmy ➜ /workspaces/RFC-3D/foo2 (mku-tshark) $ for f in witness*; do echo $f && hexdump $f; done
witness.0.POS.VxlanValidateVxlanHeader.dat
0000000 0808 0808 0808                         
0000006
witness.1.POS.VxlanValidateVxlanHeader.dat
0000000 0f0f 0f00 0f0f                         
0000006
witness.2.POS.VxlanValidateVxlanHeader.dat
0000000 0018 0000 0000                         
0000006
witness.3.NEG.VxlanValidateVxlanHeader.dat <- payload same as 0.Pos
0000000 0808 0808 0808                         
0000006
witness.4.NEG.VxlanValidateVxlanHeader.dat
0000000 0000                                   
0000001
witness.5.NEG.VxlanValidateVxlanHeader.dat
0000000 0001                                   
0000001
```

PR #114, opened 2023-11-28
EverParse3d.InputStream.Extern.fst: fix proof regression

mtzguido:
This proof regressed again. I tried to stabilize it by splittin queries, but ran into this:
```
(EverParse3d.InputStream.Extern.fst(246,1-271,5))	Query-stats (EverParse3d.InputStream.Extern.peep, 63)	failed {reason-unknown=unknown because Overflow encountered when expanding old_vector} in 42 milliseconds with fuel 0 and ifuel 1 and rlimit 256 
```
which is a z3 bug. To work around it I also added --z3refresh, and now this works reliably, albeit slowly.

I wanted to add hints but noticed there are none in the repo?

### DIFF
--- a/src/3d/prelude/extern/EverParse3d.InputStream.Extern.fst
+++ b/src/3d/prelude/extern/EverParse3d.InputStream.Extern.fst
@@ -219,7 +219,7 @@ let read
 
 #pop-options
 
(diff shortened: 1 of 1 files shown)


nikswamy:
We have not been using hints in EverParse. The proofs have generally been stable without them.
But this particular proof is a repeat offender. 
@tahina-pro can you take a look at this proof and see if it can be made more explicit/split into smaller parts etc.?

tahina-pro:
> But this particular proof is a repeat offender. @tahina-pro can you take a look at this proof and see if it can be made more explicit/split into smaller parts etc.?

I claim #115 should make this proof much more robust.

PR #116, opened 2023-12-01
3d: allow fields that consume all input

tahina-pro:
This PR allows adding a field at the end of a `typedef struct` definition to consume the remainder of the input:

```
typedef struct _test {
   UINT32 my_field;
   UINT8 remainder[:consume-all];
} test;
```

The `remainder` field uses the `parse_all_bytes` parser, which does not have the strong prefix property, so this field must be at the end of the struct.

Similarly, such a field makes its enclosing type `test` lose the strong prefix property, so the type itself must be used either at the end of a struct, or with a `[:byte-size n]`, `[:byte-size-single-element-array n]` or `[:byte-size-single-element-array-at-most n]` container.


src/3d/Ast.fst
```diff
@@ -774,6 +775,7 @@ let mk_prim_t x = with_dummy_range (Type_app (with_dummy_range (to_ident' x)) Ki
 let tbool = mk_prim_t "Bool"
 let tunit = mk_prim_t "unit"
 let tuint8 = mk_prim_t "UINT8"
+let tuint8be = mk_prim_t "UINT8BE"
```
nikswamy, line 778: Unrelated?

PR #129, opened 2024-01-30
Generate 3d probe-and-validate wrapper functions

tahina-pro:
With this PR, 3D can now generate, for a given type `T`, an unverified wrapper calling a probing function `Probe` on an address `a` and then calling the validator for `T` on the memory area probed by `MyProbe`.

If for instance in `MyModule.3d` the user writes:
 
```
extern probe MyProbe

entrypoint probe MyProbe 8
typedef struct _T {
  ... 
} T;
```
 
Then EverParse/3D generates into `MyModuleWrapper.c` and `MyModuleWrapper.h`, in addition to the ordinary `MyModuleCheckT` validator wrapper function, a probe-and-validate wrapper function:
```
boolean MyModuleMyProbeCheckT (EVERPARSE_COPY_BUFFER_T probeDest, UINT64 probeAddr);
```
 
This probe-and-validate wrapper function takes:
* a destination buffer, `probeDest`, that still needs to be provided by the caller
* the address value `probeAddr`, directly (so no need for an extra “fake input buffer” indirection)

When called, the probe-and-validate wrapper function calls the `f` probing function and, if probing succeeds, calls the `T` validator on the probed buffer.

The syntax of the `entrypoint probe` attribute is:

```
entrypoint probe probeFunctionName size
```

where the `size` argument, giving the number of bytes to probe, is either an integer literal or a global integer variable defined with `#define`. The value of `size` must be at most a 32-bit unsigned integer.


nikswamy:
Thanks Tahina! A few things we discussed:


* [ ] Let's make it `(length=8)`, so that it's more explicit and matches the existing probe syntax.
* [ ] It would also be nice if instead of a constant or defined symbol, it could also be sizeof(T)
* [ ] The name of the wrapper should be MyModule[f]ValidateT, typically this would produce `MyModuleProbeAndCopyValidateT`
* [ ] Aside from these cosmetic things, it would be great if the generated .h file only includes the probing version of the entrypoint, since the non-probing one would be unsafe to call with a user pointer.

Issue #143, opened 2024-08-19
Add offset checks when checking that 3D type refines a C type


nikswamy:
Cf. https://project-everest.github.io/everparse/3d-lang.html#checking-3d-types-for-correspondence-with-existing-c-types

We currently only add sizeof checks; we should also add offsetof checks.

tahina-pro:
Thanks Nik!
Right now #144 is enough for most cases, but does not seem to work with `#if` . To this end, we should:
* [ ] Generate `offsetof` for nested anonymous `struct`s and `casetype`s
* [ ] Make the generated `StaticAssertions.c` file include the `.h` file provided by the user (in the configuration file given as `--config`) where compilation constants are meant to be defined

PR #144, opened 2024-08-20
Adding offsetof checks as static assertions when refining a C type

nikswamy:
Cf. Issue #143

### DIFF
--- a/src/3d/Options.fst
+++ b/src/3d/Options.fst
@@ -4,30 +4,8 @@ open FStar.All
 open FStar.ST
 module U8 = FStar.UInt8
 module OS = OS
-
+open Utils
 #push-options "--warn_error -272" //top-level effects are okay
-
-inline_for_extraction
-let valid_string
-  (valid: (string -> Tot bool))
-: Tot Type0
-= (s: string { valid s == true })
-
-let always_valid (_: string) : Tot bool = true
-
-let starts_with_capital (s: string) : Tot bool =
-  String.length s >= 1 &&
-  begin let first = String.sub s 0 1 in
-    String.compare first "A" >= 0 && String.compare first "Z" <= 0
-  end
-
-let ends_with (s:string) (suffix:string) : bool =
-  let l = String.length s in
-  let sl = String.length suffix in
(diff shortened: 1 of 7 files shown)


src/3d/Options.fst
```diff
@@ -4,31 +4,9 @@ open FStar.All
...
-  end
-
-let ends_with (s:string) (suffix:string) : bool =
-  let l = String.length s in
-  let sl = String.length suffix in
-  if sl > l || sl = 0
-  then false
-  else let suffix' = String.sub s (l - sl) sl in
-       suffix = suffix'
-
```
nikswamy, line 31: Moved to Utils

src/3d/StaticAssertions.fst
```diff
@@ -43,36 +55,62 @@ let compute_static_asserts (benv:B.global_env)
...
+                    "No offsetof assertions for type %s because %s\ns"
+                    (ident_to_string j)
+                    msg)
+                  i.range;
+                []
+              | Inl offsets ->
+                offsets |>
+                List.collect
+                  (fun offset -> 
+                    if TypeSizes.is_alignment_field (fst offset)
```
nikswamy, line 82: Exclude assertions on fields that are added for alignment padding; they are implicit in the C type

src/3d/TypeSizes.fst
```diff
@@ -433,12 +466,29 @@ let decl_size_with_alignment (env:env_t) (d:decl)
...
+  | ExternType names -> [names.typedef_name; names.typedef_abbrev]
+
 let size_of_decls (genv:B.global_env) (senv:size_env) (ds:list decl) =
-  let env =
-    B.mk_env genv, senv in
-    // {senv with sizes = H.create 10} in
+  let env = B.mk_env genv, senv in
   let ds = List.map (decl_size_with_alignment env) ds in
+  let ge = B.global_env_of_env (fst env) in
+  ds |> List.iter (fun d ->
```
nikswamy, line 486: Update the global environment with revised type declarations after fields (may) have been added for alignment padding

PR #147, opened 2024-10-10
Update ASN1.Spec.Interpreter.fst with converters in ASN1.Base.fst

Black-Kamous:
Running `make` in `everparse/src/ASN1` gets error:

```
* Error 19 at ASN1.Spec.Interpreter.fst(64,80-64,103):
  - Could not prove termination of this recursive call
  - The SMT solver could not prove the query. Use --query_stats for more
    details.
  - See also ASN1.Spec.Interpreter.fst(61,2-94,60)

* Error 19 at ASN1.Spec.Interpreter.fst(101,23-101,48):
  - Could not prove termination of this recursive call
  - The SMT solver could not prove the query. Use --query_stats for more
    details.
  - See also ASN1.Spec.Interpreter.fst(97,2-101,125)
```
Thank you for replying @nikswamy .


Black-Kamous:
@microsoft-github-policy-service agree

nikswamy:
Thank you so much! I'll take a look at this over the next few days and hopefully have something to report early next week.

Issue #125, opened 2024-01-19
Optimizing parsers for arrays of constant-size elements without actions


nikswamy:
If one writes `UINT32 t[:byte-size n]`, we actually generate a while-loop to validate each UINT32 advancing by 4 bytes each time. This is inefficient, since one could instead just check that `n % 4 == 0` and skip over the array.

We already have an optimized combinator for integer types without refinements or actions, see
EverParse3d.Actions.Base.validate_nlist_constant_size_without_actions, but the interpreter does not seem to use it.

We should enhance the 3D interpreter to use the computed kinds and target this optimized combinator, i.e., relying on parser_kind_is_constant_size in TranslateForInterpreter.fst

PR #151, opened 2024-10-18
Optimize validation of arrays

nikswamy:
In Issue #125, we note that validation for arrays like `T f[:byte-size n]` should be optimized in case:

* `T` is of a fixed size `k` that divides `n`
* Validation of `T` always succeeds and does not involve any actions

instead of generating a loop to validate the array element-wise, the entire array can be checked with a single arithmetic check.

This PR achieves this by first:

1. Enhancing the indexing structure of validators so that we can statically detect from the type of a validator whether or not it runs any actions
2. Revising the EverParse interpreter to target the `validate_nlist_constant_size_without_actions` combinator, in case the element type of the array has no actions.
3. Revising the EverParse frontend to produce the appropriate metadata, both to record whether or not a combinator has actions and whether or not the size of `T` is a compile-time constant (in which case, the modulus check can be statically evaluated way).

Along the way, I migrated LowParse and 3D to use the F* option `--ext context_pruning` and simplified away the use of various settings, including the use of `--using_facts_from` in generated 3D code.


doc/3d-snapshot/GetFieldPtr.c
```diff
@@ -22,70 +22,17 @@ GetFieldPtrValidateT(
 )
 {
   /* Validating field f1 */
-  BOOLEAN hasEnoughBytes0 = (uint64_t)(uint32_t)10U <= (InputLength - StartPosition);
-  uint64_t positionAfterT;
-  if (!hasEnoughBytes0)
+  BOOLEAN hasBytes0 = (uint64_t)10U <= (InputLength - StartPosition);
```
nikswamy, line 25: NB: This is generated code from this PR branch. The while loops are gone in favor of a simple arithmetic check

nikswamy:
@tahina-pro, I merged in your PR for pairs and also for fixed size arrays (fixing one small bug, see 688122f)

But, still I don't see tuples like { UINT8 fst; UINT8 snd[7]} getting optimized into a single check

nikswamy:
Ah, I don't think coalescing checking a pair `{UINT32 fst; UINT8 f[8]}` is going to work in the current design, since the parser kind `kind_nlist` is defined as 

```

PR #162, opened 2025-01-08
Removing need for FSTAR_HOME

mtzguido:
(from commit messages:)

- F\* discovery: use FSTAR_EXE ?= fstar.exe
Note: 3d itself also inspects this variable (it was inspecting
FSTAR_HOME before). It may be easier to remove this behavior and add a
flag, so the fstar used is explicit in the Makefiles.

- Remove most uses of FSTAR_HOME
A few remain in package.sh, but there it does in fact need a repo
(afaict) so I kept it. There is also the warning in everest.sh, but I'm
not sure why it's checking for FSTAR_HOME in the first place.

  Most of this is untested... it does build, but I didn't run the workflows.

Note:  For this PR to work with FSTAR_EXE alone (and no fstar.exe in the PATH), krml must also respect FSTAR_EXE (https://github.com/FStarLang/karamel/pull/512). Or we would need a krml option to choose an fstar bypassing the detection.


mtzguido:
Updated with the `FSTAR_EXE ?= fstar.exe` convention. We pass `-fstar $(FSTAR_EXE)` to krml when needed, and 3D also now has a `--fstar` option to override the F* used (otherwise we look in the PATH).

A run of the Github actions CI going on here https://github.com/mtzguido/everparse/actions/runs/12719269191.

PR #165, opened 2025-01-13
actions: add package workflow

mtzguido:
This adds a fully-cloud pacakge job. The branches (and repos) for F* and karamel can be set from the dispatch menu.

Linux-only for now while we figure out what's going on with the stuck Windows builds for F*.


mtzguido:
Example run going on now: https://github.com/mtzguido/everparse/actions/runs/12753699154.

For faster iteration we could cache F* and karamel, and even install ccache so Z3 recompilation is almost instant.

mtzguido:
Finally added a working Windows build. Here is an [example run](https://github.com/project-everest/everparse/actions/runs/12856512953) (the commit hash is different since I rebased, but it's exactly equal to 9203ca87f7549d6678c3d4428fb898809b475db9)

PR #166, opened 2025-01-21
Replace Windows CI workflow with #165

tahina-pro:
The old Windows CI workflow was relying on a full build of F* from scratch, which is (currently) not possible on Windows. So, we need to disable it, and to replace it with the workflow that @mtzguido introduced in #165 and that builds a F* source package on Linux (or downloads a cached one) and then compiles that source package on Windows with EverParse.


.github/workflows/package-windows.yml
```diff
@@ -154,12 +154,34 @@ jobs:
...
+      - name: Download and extract artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: everparse
+          path: ${{ github.workspace }}\test
+      - name: Check out repo
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: src\3d\tests
+          path: ${{ github.workspace }}\everparse
```
mtzguido, line 183: I think you may not need all the `github.workspace` mentions here, it should be the directory where the runner is currently at. This is not so in the self-hosted runner only due to using a `--login` shell, I think.

tahina-pro: Thanks for your remark. However, I remember having issues with relative paths, so I inserted those absolute paths in a47be5ea0a7852709c6ef171ce8ca783ed226ba0.

PR #182, opened 2025-04-04
Using pretty names for the bijections

mtzguido:
During compilation, we try to detect the names of elements in maps and arrays and a `named` tag (from the Prettify tactic) to the types, so the tactic uses these names as the field names in the constructors, which are then used by karamel to choose the strucuture field names.

This PR bumps F* with a few patches over master-- the prettify tactic generates a bijection that does not extract, I disabled this in the patch. I'd like to instead toggle this externally. An `inline_for_extraction` qualifier is also needed for `named`, but that does not really belong in that module.

The names for anonymous fields is `_x0`, `_x1`, ..., where the count increases for any anonymous field. Maybe this should be something more prominent like `_anonN`?


EverParse.fst.config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "fstar_exe": "fstar.exe",
+    "fstar_exe": "./opt/FStar/bin/fstar.exe",
```
gebner, line 2: This makes it impossible to run everparse without the bundled F*.

gebner:
Does `intkey` also work for negative integers?

> The names for anonymous fields is `_x0`, `_x1`, ..., where the count increases for any anonymous field. Maybe this should be something more prominent like `_anonN`?

I think `_x0` is fine, and `_anon0` isn't any better.  If we want to spend any extra effort on this, we should add support for specifying names in the CDDL.

PR #187, opened 2025-06-03
CI: restore automatic doc refresh

tahina-pro:
#162 inadvertently removed the automatic refresh of the EverParse documentation to https://github.com/project-everest/project-everest.github.io/tree/master/everparse .

This PR restores it. In particular, we do not need to install Sphinx as part of CI, we install it as a Python virtual environment within `make -C doc html` instead.

This PR also removes all remnants of the old Docker-based CI.

The CI run on this branch, https://github.com/project-everest/everparse/actions/runs/15425056828/job/43409779097, pushed new docs as project-everest/project-everest.github.io@355f6878acee729095629152a048f6aab93f1a2f


doc/3d-lang.rst
nikswamy, line 1: Thanks!

PR #192, opened 2025-06-13
No PpxDerivingShow for AST

mtzguido:
F* extraction fixes will mean that the index of ast0_wf_typ becomes an Obj.t, and we cannot use deriving show on it.


mtzguido:
This made it to `main` already.

PR #195, opened 2025-07-15
EverCDDL: generalize map group footprint; remove submodules

tahina-pro:
In [Version 1 of the EverCDDL paper](https://arxiv.org/abs/2505.17335v1), the last paragraph of Section 4.1 describes a limitation on CDDL map group elaboration: footprints excluded from a table only cover map entry keys, and thus the pattern `? (18 => bstr), * (int => any)` was not supported, leading us to add cuts: `? (18 : bstr), * (int => any)`, yielding a different semantics.

This PR lifts this limitation: now, footprints cover both map entry keys and values. Thus, this PR now enables support for `? (18 => bstr), * (int => any)` . By contrast, `(18 => bstr // 42 => int), * (int => any)` is not supported and for now the user has to rewrite the CDDL spec by hand to `(18 => bstr, * int => any) // (42 => int, * int => any)`. This PR rewrites COSE accordingly.

This PR also removes submodules so that all master/main branches for F*, Karamel and Pulse are used by default.


tahina-pro:
Old vs. new benchmark results:
[20250715diff.txt](https://github.com/user-attachments/files/21249933/20250715diff.txt)

mtzguido (approved):
For the CI, modules, and benchmarks: looks good to me.

PR #204, opened 2025-07-24
test: Allow to disable diffing with EVERPARSE_NO_DIFF=1

mtzguido:


mtzguido:
This needs a snapshot update to be green. I don't know which rule to run for that.

mtzguido:
Everest failed again due to this. I'll merge.

PR #205, opened 2025-07-24
Remove HACL* from opt/

tahina-pro:
This PR removes HACL* from ` opt/`, since it is not needed to build EverCOSE. Rather, this PR makes HACL* local to the COSE verified interop test: just like qcbor and tinycbor, HACL* is now downloaded by that test, and only then; and it is skipped if `HACL_HOME` is set.

### DIFF
--- a/.github/workflows/cbor.yml
+++ b/.github/workflows/cbor.yml
@@ -1,0 +1,17 @@
+name: Test CBOR, COSE without F*
+on:
+  push:
+    branches-ignore:
+    - _**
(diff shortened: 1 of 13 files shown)


tahina-pro:
This PR creates new workflows that I want to test with workflow_dispatch instead of having to push and reset the 2-hour-long PR CI task. Thus I need to have them merged into `master` as empty shell first, because they need to be in `master` for `workflow_dispatch`  to run.

PR #218, opened 2025-08-11
Update 3d snapshot

mtzguido:
Karamel changed the shape of inclusion guards.
https://github.com/FStarLang/karamel/issues/637


mtzguido:
This is missing updates to the other snaphots. Closing and running the workflow instead.

PR #223, opened 2025-08-13
Fix COSE interop segfault on aarch64.

gebner:


tahina-pro:
Thanks Gabriel! I successfully tested in an amd64 Ubuntu 22.04 and 24.04 Docker image. I now need to test in an Apple Silicon Ubuntu 24.04 Docker image.

PR #227, opened 2025-08-24
Use a standalone opam root, fuse opam package build into `make -C opt`

tahina-pro:
Currently, the user needs to run `make _opam && eval $(opam env) && make -C opt` to build EverParse dependencies. Moreover, `make _opam` touches the user's `.opam` by registering a switch.

This PR fixes all of those issues. Now, the user only needs to run `make -C opt` to build all EverParse dependencies at once, including opam packages. To perform the latter, `make -C opt` now creates a standalone opam root in `opt/opam`.

Moreover, `make -C opt` now creates `opt/opam-env.Makefile`, which is automatically included by the main `Makefile` to set up the opam environment in addition to `FSTAR_EXE`, etc. when invoking one of its rules. Thus, the user no longer needs to run `eval $(opam env)`.

As before, if the user needs to work beyond the main Makefile rules, a single invocation of `eval $(make -s env)` is enough to set up both opam and EverParse environments at once.

### DIFF
--- a/.github/workflows/cbor.yml
+++ b/.github/workflows/cbor.yml
@@ -16,6 +16,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
         with:
+          submodules: true
           path: everparse
       - run: sudo apt update && sudo apt install -y libssl-dev
       - name: Everparse CI
@@ -32,6 +33,7 @@ jobs:
           ocaml-compiler: 5
       - uses: actions/checkout@master
         with:
+          submodules: true
           path: everparse
       - name: Everparse CI
         run: |
@@ -49,6 +51,7 @@ jobs:
           brew install bash gnu-getopt make gnu-time
       - uses: actions/checkout@master
         with:
+          submodules: true
           path: everparse
       - name: Everparse CI
         run: |
@@ -64,6 +67,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
         with:
+          submodules: true
           path: everparse
       - name: Havoc .depend
         run: |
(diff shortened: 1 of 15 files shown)


tahina-pro:
Now, with this PR, ˋmake -C opt` will automatically build opam package dependencies under `opt/opam`, unless `EVERPARSE_USE_OPAMROOT` is set, in which case the user's existing opam setup instead.

mtzguido:
Here's my (opinionated) 2 cents, some of which we discussed today:

If everparse has local copies of F*/karamel/pulse (via submodules, hashes, or whatever), then it should just use them, and not look into the user's environment at all (e.g. reading `*_HOME` variables, or trying to use an F* from the PATH). I routinely have to tweak things since everparse tries to call `$FSTAR_EXE` (which I have unset), or because it read my `KRML_HOME` instead of using the one in opt/.

I think the default should be to just use the dependencies in opt, including for CI purposes. This way, there is no leeway in what is being checked: we check that everparse builds and passes the tests with the checked-in snapshots (submodules/hashes) of the projects. This also means that changes in F*/krml/pulse will not break the everparse master branch, as there is no implicit requirement of being up-to-date.

Of course, we want to be up-to-date anyway, so a nightly job could try to advance the submodules, test the build, and commit this new state if everything passes.

If a user wants to use a local F*/krml/Pulse, this should be a very explicit step to deviate from the norm. Maybe running `make use-local` or similar would 1) write a config file with pointers to the system F*/krml/pulse, 2) disable the automatic building of opt/. One can also add a top-level `fstar.sh` script to either call the F* in opt, or the system one according to this setting, and use this script as the fstar_exe in the VS Code configs, to make this work nicely in the IDE too.

This is what I do for Kuiper (except for allowing using external dependencies, I haven't seen a need for this yet) and it's been working nicely.

PR #232, opened 2025-09-12
New auto-configurable Makefile and release system

tahina-pro:
TL;DR: This PR:
* updates the Makefile to build EverParse together with all of its dependencies automatically, with support for customization via environment variables
* creates a new push-button release system to push native Windows, MacOS and Linux binary packages into GitHub releases and a Docker image into GitHub Packages
* updates the documentation accordingly

This PR subsumes #227 

# New Makefile

With this PR, a user just needs to type `make` and everything will be built: opam packages, F*, Karamel, Pulse, with Z3 automatically downloaded, and all paths set up. Under the hood:

1. The Makefile ~downloads the Everest project into `opt/everest` (unless already present),~ initializes an opam root in `opt/opam` with OCaml 5.3.0, and ~calls `opt/everest/everest opam`~ collects the opam package definitions of F*, Karamel and Pulse to install the opam package dependencies. (This ~comes~ inspires from #227)
  The user can supersede this whole step with their own opam installation (or `OPAMROOT`) by setting `EVERPARSE_USE_OPAMROOT=1`. Then, it becomes the user's responsibility to install the opam packages.
2. Then, the Makefile creates and includes `opam-env.Makefile` to properly set up the environment variables to OCaml and opam (including `OCAMLPATH`, `PATH`, etc.)
3. If ~`FSTAR_EXE` is not set~ `EVERPARSE_USE_FSTAR_EXE` is not set to 1, then the Makefile downloads F* into `opt/FStar` (unless already present), compiles it and sets `FSTAR_EXE` to `opt/FStar/out/bin/fstar.exe`
  If the user sets ~`FSTAR_EXE`~ `EVERPARSE_USE_FSTAR_EXE=1`, then `EVERPARSE_USE_OPAMROOT=1` is automatically set because EverCDDL has a plugin that needs to be compiled with the very same OCaml setup as the one compiling F*. (Thus, users can no longer use F* binary packages to build EverParse.)
  If the user sets `EVERPARSE_USE_FSTAR_EXE=1` without setting `FSTAR_EXE`, then `FSTAR_EXE=fstar.exe` by default (relying on `PATH`)
4. If ~`KRML_HOME` is not set~ `EVERPARSE_USE_KRML_HOME` is not set to 1, then the Makefile downloads Karamel into `opt/karamel` (unless already present), compiles it and sets `KRML_HOME` to `opt/karamel`
  If the user sets ~`KRML_HOME`, then `FSTAR_EXE` is required~ `EVERPARSE_USE_KRML_HOME=1`, then automatically `EVERPARSE_USE_FSTAR_EXE=1`, because the `.checked` files in Karamel must match the user's F* install, and EverParse is not going to recheck Karamel in that case.
5. If ~`PULSE_HOME` is not set~ `EVERPARSE_USE_PULSE_HOME` is not set to 1, then the Makefile downloads Pulse into `opt/pulse` (unless already present), compiles it and sets `PULSE_HOME` to `opt/pulse/out`
  If the user sets ~`PULSE_HOME`, then `FSTAR_EXE` is required~ `EVERPARSE_USE_PULSE_HOME=1`, then automatically `EVERPARSE_USE_FSTAR_EXE=1`, because the `.checked` files in Pulse must match the user's F* install, and EverParse is not going to recheck Pulse in that case.
6. If Z3 4.13.3 cannot be found via `fstar.exe --locate_z3`, or in the PATH, or in `opt/z3`, then the Makefile downloads it into `opt/z3` and updates the `PATH` accordingly.
  This step is skipped if `ADMIT=1` is provided (which automatically appends `--admit_smt_queries true` to `OTHERFLAGS`.)
7. Then, the Makefile creates and includes the global `.depend` F* dependency file as done currently. This dependency file itself depends on F*, Karamel and Pulse as determined with internal `NEED_OPAM`, `NEED_FSTAR`, `NEED_KRML`, `NEED_PULSE` and `NEED_Z3` variables automatically computed from the presence or absence of the corresponding user environment variables at the beginning of the Makefile.
8. Then, the rest of the build happens as before, since all auxiliary Makefiles have been included at this point.

Steps 1 to 6 can be reproduced with `make deps`.

The only dependencies not automatically built are Rust and system packages that need to be installed as root. For transparency reasons, we describe them explicitly in the README.md, rather than letting `opam depext` automatically install them behind the user's back (also because there is no way for opam to ask for confirmation for `opam depext` but not for the rest of `opam install`, since we want to skip any user interaction when setting up anything in `opt/`)

Then, the user can automatically set up their Bash environment with `eval "$(make -s env)"` . This will set up the OCaml and opam environments, `FSTAR_EXE`, `KRML_HOME`, `PULSE_HOME`, `EVERPARSE_HOME` and `PATH` accordingly. For users' convenience, we provide a Bash wrapper script, `./shell.sh`, to open a Bash shell (or run a command) with the environment set up accordingly.

# New release system

This PR defines `.github/workflows/release.yml` so that we just need to push a single button to provide the whole release automatically. This workflow now only uses GitHub-hosted runners, including on Windows. Example runs can be found at https://github.com/tahina-pro/quackyducky/actions/runs/17664901098 (that run has not been fully configured with Visual Studio Code yet, however) and https://github.com/tahina-pro/quackyducky/actions/runs/17686176690 (fully configured with Visual Studio Code, but with a custom version number that I set by hand at tahina-pro/quackyducky@32bbc764ad9e9b8ebcddf9c452bec6024ad996bc and that I reverted by hand at tahina-pro/quackyducky@312d0bcbb6187934e81ea273ee6ffaf80f58b4eb)

User inputs:
* The name of the EverParse release branch (by default `_release`).
* ~The repository and branch for F*~

Then, when hitting the "release" button from a branch (called the "running branch" hereafter):
1. The GitHub workflow creates a F* source package on a Ubuntu runner. (This is necessary because F* cannot be built as is on Windows.)
2. Then, from the running branch, the GitHub workflow creates a release-specific branch, `_dzomo_release_NNNN` where `NNNN` is the run ID (the number following `actions/runs` in the run URL.) This branch merges the release branch, commits the contents of the F* source package, and ~creates submodules for Karamel, Pulse, Everest and opam-repository (which allows~ uses `opam lock` to freeze the versions of the opam packages used to build everything ~)~ into `opt/`. Then this branch creates a version number of the form `vYYYY.MM.DD` and writes it into `version.txt` . Please note that the hashes of F* (from which the GitHub workflow creates a source package), Karamel and Pulse are already assumed to be recorded in the running branch, usually via the `nightly.yml` workflow (or `make -C opt advance`).
  The user can produce a specific `version.txt` or specific copies of ~Everest~, Karamel, etc. into `opt/` on the running branch before starting the workflow. In that case, the workflow will use those instead, but then the user needs to manually remove those copies from the release branch at the end of the release process, otherwise future releases will fail.
3. Then, from the release-specific branch, the GitHub workflow builds and tests MacOS, Windows and Linux binary packages, using `version.txt` for the version number. We also perform interoperability tests between binary packages to check that the hashes produced by 3D on one platform can be checked with `--check_hashes` on another platform. Binary packages are pushed as run artifacts.
4. In parallel with 3, the GitHub workflow builds a linux/amd64 Docker image containing a Visual Studio Code Server with the F* VSCode Assistant (from a04d4dd26c8bd1107d0eb0eee13a38118eb25204, thanks @gebner !), and pushes it into GitHub packages as `ghcr.io/project-everest/everparse:_dzomo_release_NNNN`. Then, a full `make test` is run in a container based on that image.
5. Once all tests succeed, the GitHub workflow pushes a new tag, `vYYYY.MM.DD`, corresponding to the version number in `version.txt`
6. Then, the GitHub workflow downloads the binary packages created as run artifacts, pushes them into a new GitHub release associated with that tag, and deletes the artifacts.
7. Then, from the release branch, the GitHub workflow merges the running branch, and then merges the release-specific branch, and then reverts that merge. At this point, the release branch and the running branch should have the same contents. This allows all release tags to be in a single ancestry lineage; thus, the message produced when creating the GitHub release can reference the immediately previous release in that lineage.
9. In parallel with 6 and 7, the GitHub workflow pushes a new Docker image tagged with that tag, based on the `:dzomo_release_NNNN` image, the only difference being that the Git clone within the Docker image is now tagged with the version tag as well.
10. Then, the GitHub workflow removes the release-specific branch.

Caveats:

* Pulse fails to build on Windows, so EverCDDL is excluded from Windows builds.
* I don't know how to refer to the OpenSSL headers on MacOS or Windows, so EverCOSign is excluded from MacOS and Windows builds. (For Windows, I think we should use SymCrypt instead of OpenSSL anyway, and I don't know how to do that either.)

# Documentation

This PR updates the README.md with all build instructions.

Since dependency build has become much easier for the user, this PR removes the whole "build" chapter from the manual.

This PR also updates the Visual Studio Code configuration files: now, the user is assumed to call Visual Studio Code from a shell suitably configured with `./shell.sh`

### DIFF
--- a/deps.Makefile
+++ b/deps.Makefile
@@ -31,7 +31,7 @@ endif
 NEED_KRML :=
 ifneq (1,$(EVERPARSE_USE_KRML_HOME))
 export KRML_HOME := $(EVERPARSE_OPT_PATH)/karamel
-NEED_KRML := $(KRML_HOME)/_build/default/src/Karamel.exe
+NEED_KRML := $(EVERPARSE_OPT_PATH)/karamel.done
 else
 export EVERPARSE_USE_FSTAR_EXE:=1
 ifeq (,$(KRML_HOME))
@@ -44,7 +44,7 @@ NEED_PULSE :=
 ifeq (,$(NO_PULSE))
 ifneq (1,$(EVERPARSE_USE_PULSE_HOME))
 export PULSE_HOME := $(EVERPARSE_OPT_PATH)/pulse/out
-NEED_PULSE := $(PULSE_HOME)
+NEED_PULSE := $(EVERPARSE_OPT_PATH)/pulse.done
 else
 export EVERPARSE_USE_FSTAR_EXE:=1
 ifeq (,$(PULSE_HOME))
@@ -56,7 +56,7 @@ endif
 NEED_FSTAR :=
 ifneq (1,$(EVERPARSE_USE_FSTAR_EXE))
 export FSTAR_EXE := $(EVERPARSE_OPT_PATH)/FStar/out/bin/fstar.exe
-NEED_FSTAR := $(FSTAR_EXE)
+NEED_FSTAR := $(EVERPARSE_OPT_PATH)/FStar.done
 z3_exe := $(shell $(FSTAR_EXE) --locate_z3 \$(Z3_VERSION) 2>/dev/null)
 ifneq (0,$(.SHELLSTATUS))
 z3_exe :=
@@ -140,8 +140,9 @@ $(EVERPARSE_OPT_PATH)/pulse/Makefile: $(EVERPARSE_OPT_PATH)/hashes.Makefile
 $(EVERPARSE_OPT_PATH)/opam.done: $(EVERPARSE_OPT_PATH)/opam/opam-init/init.sh $(EVERPARSE_OPT_PATH)/FStar/Makefile $(EVERPARSE_OPT_PATH)/karamel/Makefile $(EVERPARSE_OPT_PATH)/pulse/Makefile
 	+$(MAKE) -C $(EVERPARSE_OPT_PATH) opam.done
 
-$(EVERPARSE_OPT_PATH)/FStar/out/bin/fstar.exe: $(EVERPARSE_OPT_PATH)/FStar/Makefile $(NEED_OPAM)
-	+$(with_opam) $(MAKE) -C $(dir $<) ADMIT=1
+$(EVERPARSE_OPT_PATH)/FStar.done: $(EVERPARSE_OPT_PATH)/FStar/Makefile $(NEED_OPAM)
+	rm -f $@
+	+$(with_opam) $(MAKE) -C $(EVERPARSE_OPT_PATH)/FStar ADMIT=1
 	touch $@
 
 $(EVERPARSE_OPT_PATH)/z3: $(EVERPARSE_OPT_PATH)/FStar/Makefile
@@ -152,12 +153,14 @@ $(EVERPARSE_OPT_PATH)/z3: $(EVERPARSE_OPT_PATH)/FStar/Makefile
 	mv $@.tmp $@
 	touch $@
 
-$(EVERPARSE_OPT_PATH)/karamel/_build/default/src/Karamel.exe: $(EVERPARSE_OPT_PATH)/karamel/Makefile $(NEED_FSTAR) $(NEED_OPAM)
-	+$(with_opam) env OTHERFLAGS='--admit_smt_queries true' $(MAKE) -C $(dir $<)
+$(EVERPARSE_OPT_PATH)/karamel.done: $(EVERPARSE_OPT_PATH)/karamel/Makefile $(NEED_FSTAR) $(NEED_OPAM)
+	rm -f $@
+	+$(with_opam) env OTHERFLAGS='--admit_smt_queries true' $(MAKE) -C $(EVERPARSE_OPT_PATH)/karamel
 	touch $@
 
-$(EVERPARSE_OPT_PATH)/pulse/out: $(EVERPARSE_OPT_PATH)/pulse/Makefile $(NEED_FSTAR) $(NEED_OPAM)
-	+$(with_opam) $(MAKE) -C $(dir $<) ADMIT=1
+$(EVERPARSE_OPT_PATH)/pulse.done: $(EVERPARSE_OPT_PATH)/pulse/Makefile $(NEED_FSTAR) $(NEED_OPAM)
+	rm -f $@
+	+$(with_opam) $(MAKE) -C $(EVERPARSE_OPT_PATH)/pulse ADMIT=1
 	touch $@
 
 env:
--- a/opt/.gitignore
+++ b/opt/.gitignore
@@ -9,3 +9,6 @@ fstar-deps.opam
 karamel-deps.opam
 pulse-deps.opam
 opam.done
+FStar.done
+karamel.done
+pulse.done
--- a/opt/Makefile
+++ b/opt/Makefile
@@ -93,18 +93,18 @@ release:
 
 .PHONY: release
 
-ifeq (,$(do_not_clean_FStar))
 FStar.clean:
+	rm -f FStar.done
+ifeq (,$(do_not_clean_FStar))
 	rm -rf FStar
 else
-FStar.clean:
 	+if test -d FStar ; then $(MAKE) -C FStar clean ; fi
 endif
 
 clean_rules := $(addsuffix .clean,karamel pulse z3 opam)
 
 $(clean_rules): %.clean:
-	rm -rf $(basename $@)
+	rm -rf $(basename $@) $(basename $@).done
 
 .PHONY: %.clean
 


mtzguido:
Writing some thoughts below.

> TL;DR: This PR:
> 
> * updates the Makefile to build EverParse together with all of its dependencies automatically, with support for customization via environment variables
> * creates a new push-button release system to push native Windows, MacOS and Linux binary packages into GitHub releases and a Docker image into GitHub Packages
> * updates the documentation accordingly
> 
> This PR subsumes #227
> 
> # New Makefile
> With this PR, a user just needs to type `make` and everything will be built: opam packages, F*, Karamel, Pulse, with Z3 automatically downloaded, and all paths set up. Under the hood:
> 
> 1. The Makefile downloads the Everest project into `opt/everest` (unless already present), initializes an opam root in `opt/opam` with OCaml 5.3.0, and calls `opt/everest/everest opam` to install the opam package dependencies. (This comes from [Use a standalone opam root, fuse opam package build into `make -C opt` #227](https://github.com/project-everest/everparse/pull/227))

This seems really excessive to me for users who already have F*. It's good for archival purposes, but needlessly complex otherwise. I'll set the variable below in my shell.

>    The user can supersede this whole step with their own opam installation (or `OPAMROOT`) by setting `EVERPARSE_USE_OPAMROOT=1`.
> 2. Then, the Makefile creates and includes `opam-env.Makefile` to properly set up the environment variables to OCaml and opam (including `OCAMLPATH`, `PATH`, etc.)
> 3. If `FSTAR_EXE` is not set, then the Makefile downloads F* into `opt/FStar` (unless already present), compiles it and sets `FSTAR_EXE` to `opt/FStar/out/bin/fstar.exe`

I have `fstar.exe` in my PATH, and `FSTAR_EXE` unset. It's unexpected to me that I have to set `FSTAR_EXE=fstar.exe` for the Makefiles to use it. It's also different from the convention in https://github.com/FStarLang/FStar/discussions/3655#discussioncomment-11755993. In general, I would expect a project I download and try to build to
1) use external dependencies, and error out if it can't find them
2) unconditionally use the internal dependencies, unless some *very specific* environment variable is set, or some other way of opting out. I don't think FSTAR_EXE, KRML_HOME, etc, are those variables, I have some of them set already, but not because I intend to run some mixed build here.

>    If the user sets `FSTAR_EXE`, then `EVERPARSE_USE_OPAMROOT=1` is automatically set because EverCDDL has a plugin that needs to be compiled with the very same OCaml setup as the one compiling F*. (Thus, users can no longer use F* binary packages to build EverParse.)
> 4. If `KRML_HOME` is not set, then the Makefile downloads Karamel into `opt/karamel` (unless already present), compiles it and sets `KRML_HOME` to `opt/karamel`
>    If the user sets `KRML_HOME`, then `FSTAR_EXE` is required, because the `.checked` files in Karamel must match the user's F* install, and EverParse is not going to recheck Karamel in that case.

This seems fine, but we should make karamel installable (i.e. callable as `krml` without needing a `KRML_HOME`). If so, I would also expect it to be taken from the PATH.

> 5. If `PULSE_HOME` is not set, then the Makefile downloads Pulse into `opt/pulse` (unless already present), compiles it and sets `PULSE_HOME` to `opt/pulse/out`
>    If the user sets `PULSE_HOME`, then `FSTAR_EXE` is required, because the `.checked` files in Pulse must match the user's F* install, and EverParse is not going to recheck Pulse in that case.
> 6. If Z3 4.13.3 cannot be found via `fstar.exe --locate_z3`, or in the PATH, or in `opt/z3`, then the Makefile downloads it into `opt/z3` and updates the `PATH` accordingly.
>    This step is skipped if `ADMIT=1` is provided (which automatically appends `--admit_smt_queries true` to `OTHERFLAGS`.)

Automatically downloading stuff is weird to me, even slightly suspicious. I would make the Makefile error out and provide message with a rule that can be called to do the downloading.

> 7. Then, the Makefile creates and includes the global `.depend` F* dependency file as done currently. This dependency file itself depends on F*, Karamel and Pulse as determined with internal `NEED_OPAM`, `NEED_FSTAR`, `NEED_KRML`, `NEED_PULSE` and `NEED_Z3` variables automatically computed from the presence or absence of the corresponding user environment variables at the beginning of the Makefile.
> 8. Then, the rest of the build happens as before, since all auxiliary Makefiles have been included at this point.
> 
> Steps 1 to 6 can be reproduced with `make deps`.
> 
> The only dependencies not automatically built are Rust and system packages that need to be installed as root. For transparency reasons, we describe them explicitly in the README.md, rather than letting `opam depext` automatically install them behind the user's back (also because there is no way for opam to ask for confirmation for `opam depext` but not for the rest of `opam install`, since we want to skip any user interaction when setting up anything in `opt/`)
> 
> Then, the user can automatically set up their Bash environment with `eval "$(make -s env)"` . This will set up the OCaml and opam environments, `FSTAR_EXE`, `KRML_HOME`, `PULSE_HOME`, `EVERPARSE_HOME` and `PATH` accordingly. For users' convenience, we provide a Bash wrapper script, `./shell.sh`, to open a Bash shell (or run a command) with the environment set up accordingly.

tahina-pro:
Thanks Guido, Gabriel and Nik for our discussions so far:

Here is my plan for the next candidate:
* [ ] Create a `./fstar.sh` script to run F* from VSCode config files without the developer setting up the environment explicitly
* [ ] Use `EVERPARSE_` specific _input_ environment variables to control the Makefile, rather than `FSTAR_EXE`, `KRML_HOME`, `PULSE_HOME` (but still configure the latter variables accordingly)
* [ ] Locally manage hashes for F*, Karamel and Pulse always, instead of pulling their master/main branches on master and using submodules on release
* [ ] Locally manage opam package versions with `opam lock`: write a dummy everparse.opam to take the package lists from F*, Karamel and Pulse, and additional everparse-specific opam packages, instead of drawing from the Everest script
* [ ] Write a nightly task to upgrade hashes and package versions

tahina-pro:
> 1. I haven't been able to figure out how to disable the built-in opam root.  It always runs `opam init`.

Even with `EVERPARSE_USE_OPAMROOT=1` ?

> 2. `export EVERPARSE_USE_{OPAMROOT,FSTAR_EXE,KRML_HOME,PULSE_HOME}=1` is quite a few environment variables.  (And it took me a few tries to figure out if I need to set it to 0 or 1.)  Could we trim that down to `export EVERPARSE_USE_OPT=0`?

If needed, I can add such a shortcut that will set all such variables to 1.

mtzguido:
> Even with `EVERPARSE_USE_OPAMROOT=1` ?

This did work for me FWIW. (And I think the shortcut would be nice too!)

tahina-pro:
Thanks to Guido's comments, I fixed the `make clean` rule, and I added a `make distclean` rule:
* `make clean` cleans up all of EverParse-only build products
* `make distclean` cleans up all of EverParse build products and removes everything in `opt` (opam, F*, Karamel, etc.) I tested and I can confirm that `make distclean` yields a fully clean state, even after `make test`. I haven't dared to add a CI test for that, though.

gebner:
I'm running into a new error now:
```
$ make ADMIT=1
[...]
/home/gebner.linux/everparse/opt/FStar/out/bin/fstar.exe  --z3version 4.13.3  --admit_smt_queries true --include /home/gebner.linux/everparse/opt/karamel/krmllib --include /home/gebner.linux/everparse/opt/karamel/krmllib/obj --include /home/gebner.linux/everparse/opt/pulse/out/lib/pulse --include src/lowparse --include src/ASN1 --include src/3d/prelude --include src/cbor/spec --include src/cbor/spec/raw --include src/cbor/spec/raw/everparse --include src/cddl/spec --include src/lowparse/pulse --include src/cbor/pulse --include src/cbor/pulse/raw --include src/cbor/pulse/raw/everparse --include src/cddl/pulse --include src/cddl/tool --include . --cache_checked_modules --warn_error @241 --already_cached PulseCore,Pulse,C,LowStar,*,-LowParse,-EverParse3d,-ASN1,-CBOR,-CDDL,Prims,FStar,LowStar --cmi --ext context_pruning  --dep full @.depend.rsp --output_deps_to .depend.aux
* Warning 152:
  - Not a valid include directory:
      /home/gebner.linux/everparse/opt/karamel/krmllib/obj

* Warning 274:
  - Implicitly opening namespace 'cbor.pulse.api.det.' shadows module 'c'
    in file "/home/gebner.linux/everparse/opt/karamel/krmllib/C.fst".
  - Rename
    "/home/gebner.linux/everparse/opt/karamel/krmllib/C.fst"
    to avoid conflicts.

* Error 317:
  - Expected C.Loops to be already checked but could not find it.

1 error was reported (see above)
```

tahina-pro:
This is because I am not tracking the right witness to declare Karamel correctly built. Let me fix that. Thanks Gabriel!

PR #244, opened 2025-10-30
Upgrade to F* Universes

tahina-pro:
FStarLang/FStar#3699 makes many EverParse proofs fail. This PR, authored by @nikswamy, fixes such proofs. Thanks a lot Nik!

Some remaining admits in this PR are slated to be overwritten once EverCDDL serialization is generalized beyond deterministically encoded CBOR.


opt/advance.Makefile
```diff
@@ -1,3 +1,3 @@
 FStar_hash := origin/master
-karamel_hash := origin/master
+karamel_hash := fb36fecb552c9fb202beb38a6c5a732c3f2cd49f
```
tahina-pro: 7f02f8554be72f3d00f8c970a5ea8a0612c80571 should be reverted once FStarLang/karamel#676 is solved

PR #271, opened 2026-03-27
Port LowParse combinators to Pulse

tahina-pro:
In light of the upcoming deprecation of Low*, this PR is porting LowParse combinators to Pulse.

This PR is also porting some LowParse.Pulse combinators from ACM CCS 2025 to a new separation logic predicate, `pts_to_parsed`, which relies on the parser specification rather than the serializer specification, so that (most of) those combinators can be used with non-injective parsers.

Coarse-grained roadmap:

* [x] LowParse combinators 
* [x] LowParse Low* tests (but not SLow, which are slated to disappear)
* [x] QuackyDucky
* [x] QuackyDucky tests

Left unported: offset reasoning and finalizers for serialization. These should be replaced with wholesale serializers, from our ACM CCS 2025 paper


tahina-pro:
Some accessors are missing: array_nth and the like

PR #274, opened 2026-03-28
Make filter patterns compatible with both relative and absolute paths

elefthei:
fstar --dep full now generates absolute paths in .depend files. This breaks `$(filter src/X/%,...)` patterns in EverParse Makefiles since they assume relative paths.

Fix: introduce `_EP` (EverParse Path prefix) variable that auto-detects whether `.depend` uses absolute or relative paths. If absolute, `_EP` is set to `$(CURDIR)/` so that patterns like `$(_EP)src/lowparse/%` match both formats. If relative, `_EP` is empty (backward compatible).

Updated 11 Makefiles across the build system.


tahina-pro:
Deprecated in favor of FStarLang/FStar#4156

PR #292, opened 2026-06-04
Build EverParse against a F* binary package; configure via `./configure`

tahina-pro:
## Motivation

On the `fstar2` branch, EverParse builds F* (and, transitively, Karamel and
Pulse) **from source**: `deps.Makefile` clones the F* repo under `opt/`, builds
it, downloads Z3, and threads everything together through a set of
`EVERPARSE_*` environment variables.

F* now publishes **binary packages** (the nightly tarballs at
https://github.com/FStarLang/fstar-nightly), which bundle `fstar.exe`, the
`krml` (Karamel) executable, the Z3 versions F* needs, and — as of
**FStarLang/FStar#4300** — `fstar.lib` shipped as OCaml **sources + a dune
project**, installable via `fstar.exe --install_lib_with_deps`.

This PR makes EverParse able to build and test against such a binary package
(no F* source clone, no F* build, no Z3 download), while keeping the existing
build-F*-from-source path fully working. Along the way it replaces the
`EVERPARSE_*` environment-variable configuration with a `./configure` script
that records choices in a generated `config.Makefile`.

> **Depends on FStarLang/FStar#4300**, which adds `fstar.exe --install_lib[_with_deps]`
> and makes binary packages ship `fstar.lib` as source. EverParse's
> `fstarlib.done` rule and the binary-package workflow rely on that command.

## Configuration: `./configure` + `config.Makefile` (replaces `EVERPARSE_*` env vars)

- New **`./configure`** script takes dependency options as command-line
  arguments and produces **`config.Makefile`**, which `deps.Makefile` includes
  via a prerequisite-less rule that runs `./configure` with no arguments by
  default. `src/z3-version.Makefile` also includes `config.Makefile`, so the Z3
  version can be overridden there. `package.sh`, CI sparse-checkouts,
  `.gitignore` and `README.md` are updated accordingly.
- **Decoupled `--use-fstar-exe`/`--use-krml-exe` from the opam root.** Supplying
  a pre-existing `fstar.exe`/`krml` no longer implies using the ambient opam
  switch. Users who need the ambient switch (e.g. to compile the EverCDDL F*
  plugin against their F*'s OCaml environment) pass **`--use-opamroot`**
  explicitly.
- New **`--opamroot=PATH`** option (implies `--use-opamroot`): records
  `OPAMROOT := PATH` in `config.Makefile`; `deps.Makefile` exports it so
  EverParse can target a dedicated opam root without setting `OPAMROOT` in the
  environment.

## F* binary-package shortcuts

Two `./configure` shortcuts point EverParse at a F* binary package:

- **`--fstar-bin-package=DIR`** — use an already-extracted package rooted at
  `DIR`; equivalent to `--fstar-exe=DIR/bin/fstar.exe --krml-exe=DIR/bin/krml`.
- **`--fstar-bin-package-tar=FILE`** — extract a package tarball under
  `opt/fstar-bin-package/` (deliberately not `opt/fstar`, to avoid colliding
  with `opt/FStar` on case-insensitive filesystems) and use the result.

Explicit `--fstar-exe`/`--krml-exe` still take precedence; relative paths are
resolved against the invocation directory; the extracted dir is gitignored and
removed by `distclean`.

## Installing `fstar.lib` for leaf builds

EverParse's 3d/ASN1/cddl tools link F*-extracted OCaml against `fstar.lib`. In a
binary package `fstar.lib` is shipped as source, so it must be built/installed
into the configured opam switch:

- New **`fstarlib.done`** rule runs `fstar.exe --install_lib_with_deps`
  (from FStarLang/FStar#4300) under the EverParse opam env. It depends on
  `$(NEED_FSTAR)`, so a source build rebuilds F* first; with a binary package
  `NEED_FSTAR` is empty and the rule just installs `fstar.lib`. The leaf rules
  that link against `fstar.lib` (`3d-exe`, `asn1-test`, `cddl-tool`,
  `cddl-unit-tests`) and the `deps` aggregate now depend on it.
- **Decoupled opam-root creation from the F* sources.** `opt/Makefile`'s
  `opam.done` previously installed both `everparse-deps.opam` **and**
  `fstar-deps.opam` (a copy of F*'s own opam file) — forcing a F* source clone
  just to populate the opam root. It now installs only `everparse-deps.opam`;
  F*'s OCaml build deps are installed by the from-source `FStar.done` rule.
- **Declare the OCaml deps that the F* build used to provide transitively.** In
  a source build, `menhir`/`process`/`batteries` were pulled in via the F*
  build; with a binary package they are not, so they are now declared
  explicitly in `everparse-deps.opam` (used by `qd`, `3d`, `ASN1` tools).

## Z3 in binary-package mode

The `$(FSTAR_EXE) --locate_z3` probe previously ran only in source mode. With a
binary package the probe was skipped, so Z3 was looked up via `which
z3-VERSION` and, if absent from `PATH`, **downloaded into `opt/z3` — whose rule
clones the F* sources**, defeating the purpose of the binary package. The probe
now runs in **both** modes after `FSTAR_EXE` is determined: in source mode
`fstar.exe` doesn't exist yet so it falls back to PATH/download as before; with
a binary package it returns the **bundled** Z3, so no download or F* clone
happens.

## `DICE_HOME` for the Pulse DICE examples

Binary packages don't ship the Pulse examples that the CDDL DPE tests use. A new
**`--dice-home`** configure option sets their location; `deps.Makefile` keeps
the from-source default and exports `DICE_HOME`. When `DICE_HOME` points at a
non-existent dir, `src/cddl/tests/dpe` populates it via a treeless/shallow/sparse
partial clone of `FStarLang/FStar` at the commit reported by `fstar.exe
--version`, so the examples match the F* in use.

## `OCAMLPATH` handling in subsidiary Makefiles

The subsidiary OCaml build Makefiles previously forced `fstar.lib` onto
`OCAMLPATH` via `fstar.exe --ocamlenv` wrappers. These were first removed (to let
`fstar.lib` resolve from the ambient opam switch), then **restored**, because a
**source build** keeps `fstar.lib` under `opt/FStar/out/lib` (not on
`OCAMLPATH`) — so without the wrappers `dune`/`ocamlfind` couldn't find it and
`make test` failed. The roundtrip test additionally keeps an explicit
`OCAMLPATH` entry for `evercddl.lib` (under `$EVERPARSE_HOME/lib`), which
`--ocamlenv` does not provide.

## Build hygiene & CI

- Wire the `qd/unit` and `lowparse` Pulse-test clean rules into the top-level
  `make clean` (their generated artifacts previously leaked); `distclean`
  removes the new stamps and `opt/fstar-bin-package`.
- **Disable the release workflow** by renaming `release.yml` →
  `release.yml.disabled` (only `.yml`/`.yaml` are processed as workflows;
  matches the repo's existing `.disabled` convention), while the opam/F*
  dependency wiring is reworked.

## Testing the binary-package build: `fstarpkg.Dockerfile`

New `fstarpkg.Dockerfile` builds and tests EverParse against a downloaded F*
binary package tarball (`fstar*.tar.gz`) via `--fstar-bin-package-tar`, running
`make -j$(nproc) -k lowparse` (no `ADMIT`, so it exercises the package's bundled
Z3) followed by `ADMIT=1 make -j$(nproc) -k test`. `fstar-*.tar.gz` is gitignored
so the package is not committed. README also notes that `clang` is required (the
COSE interop test Makefiles hardcode `CC := clang`).

## Files touched (binary-package work)

```
 .github/workflows/{ci,nightly,nightly-fstar2}.yml   # config.Makefile sparse-checkout
 .github/workflows/release.yml -> release.yml.disabled
 .gitignore                                          # config.Makefile, fstarlib.done, fstar-*.tar.gz, opt/fstar-bin-package
 Makefile                                            # fstarlib.done deps for leaf rules; pulse-test clean rules
 README.md                                           # ./configure usage, binary-package options, clang
 configure                                           # NEW: the configure script
 deps.Makefile                                       # config.Makefile include; fstarlib.done; Z3 probe; DICE_HOME; opam.done decoupling
 fstarpkg.Dockerfile                                 # NEW: binary-package test image
 opt/Makefile                                        # opam.done installs only everparse-deps.opam
 opt/everparse-deps.opam                             # + menhir, process, batteries
 share/everparse/tests/qd/Makefile                   # clean forwarding
 src/cddl/tests/dpe/Makefile                         # DICE_HOME partial-clone rule
 src/package/package.sh                              # config.Makefile awareness
 src/z3-version.Makefile                             # include config.Makefile (Z3 version override)
```

## How to build against a binary package

```sh
# extracted package:
./configure --fstar-bin-package=/path/to/fstar
# or an unextracted tarball:
./configure --fstar-bin-package-tar=/path/to/fstar-Linux-x86_64.tar.gz

make -j$(nproc) -k lowparse        # exercises the bundled Z3
ADMIT=1 make -j$(nproc) -k test
```

## Compatibility / before merging

- The from-source build is unchanged in behavior: `./configure` with no
  arguments reproduces the old defaults, F* is still cloned/built under `opt/`,
  Z3 is still downloaded, and `fstar.lib` resolves from `opt/FStar/out/lib`.
- **Pre-merge cleanup:** revert the temporary `(TEMP) use my F* fork` commit
  (and any hash bumps to the fork) so `opt/` points back at the upstream F*
  repo/commit; merge only after FStarLang/FStar#4300 lands and is reflected in
  `opt/hashes.Makefile`.


tahina-pro:
`fstarpkg.Dockerfile` works with the F* binary package obtained from https://github.com/tahina-pro/FStar/actions/runs/26922219139

tahina-pro:
Ultimately, it has been decided that EverParse shall build F* from its sources, or at best, from a source package, since QuackyDucky, the 3D frontend and EverCDDL all require fstar.lib, and OCaml (opam/ocamlfind) packages cannot ship as binaries.

PR #293, opened 2026-06-14
Carry validated length in size_t not uint8_t in CBOR c example (documentation)

hkscy:
Classified as a documentation issue in prior correspondence with Everparse team.

The [CBOR c example](https://github.com/project-everest/everparse/blob/6b04a113db180e703d19f77ff1235500e6336c71/src/cbor/pulse/det/c/example/main.c) stores the `size_t` result of `cbor_det_validate` in `uint8_t validated_size` (ln. 230). For objects with an encoded length of 256 or more this truncates the value, so the length passed to `cbor_det_parse` is no longer the validated length.

### DIFF
--- a/src/cbor/pulse/det/c/example/main.c
+++ b/src/cbor/pulse/det/c/example/main.c
@@ -227,7 +227,7 @@ int main(void) {
      space.
   */
   // success
-  uint8_t validated_size = cbor_det_validate(output, output_size);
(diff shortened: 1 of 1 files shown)


tahina-pro:
Thanks a lot for fixing this issue! I acknowledge that, before your PR, there was some risk that users (and agents...) would get misled by the wrong pattern.

PR #300, opened 2026-07-15
QuackyDucky: copyful parsing and serialization to/from Pulse data types (`qd -pulse`)

tahina-pro:
## Summary

This PR adds a new **`-pulse` backend to QuackyDucky** (`qd`) that generates verified
*copyful* parsers and serializers between wire-format byte slices and idiomatic,
heap-allocated, **owned Pulse data types**. It is backed by a substantial extension of
the existing `LowParse.Pulse.*` / `LowParse.PulseParse.*` combinator library.

The existing non-Pulse backends are unchanged: the default (`-high` + `-low`), `-low`,
and `-high` outputs remain **byte-for-byte identical to `upstream/master`**. All new
behavior is gated behind the `-pulse` flag.

## Motivation

QuackyDucky previously offered two backends:

- `-high` — a purely functional specification (parsers/serializers over `Seq`), and
- `-low` — zero-copy Low\* validators/jumpers/accessors that read *in place* from an
  input buffer without materializing a value.

Neither produces ordinary Pulse data structures that a caller can allocate, read,
mutate, and free. The new `-pulse` backend fills that gap with a **copyful** model:
parsing reads the wire format into a freshly allocated Pulse value; serialization
writes a Pulse value back out to a byte slice with a bounds-checked ("graceful")
failure path; and each type additionally gets a size computation and a `free`.

`upstream/master` already ships a base Pulse library (zero-copy validators, jumpers,
and leaf readers). This PR builds the copyful layer on top of it.

### 1. Copyful combinator layer in `LowParse.Pulse.*` / `LowParse.PulseParse.*`

The library gains four operations per parseable type, organized around a
representation-matching relation:

- **`vmatch : t' -> tm -> slprop`** relates a low, owned runtime representation `t'`
  to a "mid" spec value `tm`, together with a partial **`conv : tm -> GTot (option t)`**
  bridging the mid to the high spec type `t`.
- **`copyful_parse vmatch p conv`** — read a validated/parsed input slice into a freshly
  allocated owned `t'`.
- **`free_t vmatch`** — free the owned representation.
- **`l2r_safe_writer vmatch s conv`** — serialize an owned `t'` into an output slice,
  signaling out-of-space through an error flag instead of failing hard (graceful).
- **`l2r_safe_size vmatch s conv`** — compute the exact serialized size of an owned `t'`.

Combinators are provided (and generated) across the full QuackyDucky type surface:

- **Leaves**: fixed-size integers, little-endian integers, and the variable-length
  length headers (`bitcoin_varint`/bcvli, `asn1_len`/`asn1_len8` DER, bounded ints).
- **Bytes**: fixed-length and bounded variable-length byte blobs
  (`flbytes` / `vlbytes` / `vlbytes'` explicit-header / `vlgenbytes` generic-header),
  emitted **Seq-natively with no `FStar.Bytes` whatsoever** (see §4).
- **Pairs / synth**: structs and type aliases.
- **Sums**: closed sums, open sums (`dsum`), and implicit-tag sums, in both a
  *leaf-readable* form (when every case is leaf-readable) and an *owned* inductive
  form otherwise.
- **Lists / arrays**: fixed-count arrays, count-prefixed lists (`vclist`),
  byte-length-bounded lists, and `fldata`/`vlgen`-framed payloads.
- **if-then-else** types (see §3).

A pure, structural **`leaf_size`** library (`LowParse.Pulse.Base` / `Combinators` /
`Sum`) computes exact leaf sizes without any dependency on `LowParse.SLow`, so the
`-pulse` output does not pull in the SLow (`size32`) machinery at all.

New modules: `LowParse.Pulse.BoundedIntLE`, `LowParse.PulseParse.BoundedIntLE`,
`LowParse.PulseParse.SizeLeaf` (plus their `.fsti`s). Many existing Pulse modules are
substantially extended (e.g. `PulseParse.Base`, `.Sum`, `.VCList`, `.Bytes`, `.List`,
`.VLData`, `.VLGen`, `.Array`, `.FLData`, `.Enum`, `.Combinators`, `.IfThenElse`).

### 2. QuackyDucky `-pulse` code generation

For every generated type `<n>`, `qd -pulse` now emits (across `<n>.fst` / `<n>.fsti`):

- the owned low representation and its `<n>_vmatch` / `<n>_conv`,
- `read_<n>` (copyful parse) and `free_<n>`,
- `write_<n>` (graceful safe writer) and `size_<n>`,
- the parser/serializer/validator/jumper specifications,

wiring the combinator library structurally so that composite types
(structs, sums, lists, arrays, framings) are built from their components' copyful
operations. Sum dispatch uses the reducible `[@Norm]` enum destructors for
extractable key handling.

### 3. `if-then-else` types

A new discriminated field form is supported: `(if tag = "<hex>" T else F)`.

- The discriminant tag and its constant are represented **Seq-natively**
  (`Seq.lseq`/`Seq.seq`), with **no `FStar.Bytes` and no `assume`**. The tag
  comparison is a C-extractable byte-by-byte test driven by a generated `get_byte`
  callback.
- Optional **custom constructor names**: `(if tag = "<hex>" Ctor1 : T else Ctor2 : F)`;
  the bare form keeps the default `<N>_true` / `<N>_false` constructors.
- A copyful interface is emitted so an if-then-else can itself be a **struct field**.
- `bytesize_eqn` lemmas are emitted so enclosing length proofs decompose an
  if-then-else field by SMT pattern.
- **`-pulse` only for the new Seq-based codegen.** The legacy non-Pulse
  (`FStar.Bytes`-based SLow/Low\*) if-then-else backend is **kept unchanged** and is
  still selected under the default / `-low` / `-high` backends.

### 4. Fully `FStar.Bytes`-free `-pulse` output

Under `-pulse`, **all** byte types are emitted Seq-natively — over `Seq.seq FStar.UInt8.t`
/ `Seq.lseq FStar.UInt8.t` instead of `FStar.Bytes` — so the generated `-pulse` F\* **and
the extracted C are entirely `FStar.Bytes`-free**. This covers:

- fixed-length bytes (`opaque[k]`) → `Seq.lseq`;
- bounded variable-length bytes with a tight fixed-width header (`opaque<lo..hi>`) →
  `Seq.seq` via `parse_bounded_seq_vlbytes`;
- bounded bytes with a **generic / variable-width length header**
  (`bitcoin_varint` / `asn1_len` / `asn1_len8`) → `parse_bounded_seq_vlgenbytes`;
- bounded bytes with an **oversized fixed-width header** (`l > log256' max`, produced by
  staging implicit-sum payloads over `vlbytes`) → `parse_bounded_seq_vlbytes_gen`.

The supporting library gained a Seq-native byte layer that mirrors the `FStar.Bytes`
one but instantiates the byte-payload serializer with the identity
`serialize_seq_all_bytes`, so all the `B32.reveal` / `B32.length` coercions disappear:
- `LowParse.Spec.SeqBytes`: the explicit-`l` family (`parse`/`serialize`/`length_serialize`/
  `_bytes_eq` `_bounded_seq_vlbytes_gen`, over `parse_bounded_vldata_strong'`) and the
  generic-header family (`_bounded_seq_vlgenbytes`, over `parse_bounded_vlgen`).
- `LowParse.Pulse.SeqBytes`: the corresponding `validate` / `jump` / `copyful_parse` /
  `l2r_safe_writer` / `l2r_safe_size` combinators for both families (plus the primed
  vldata-strong payload copyful and the generic `vlgen` byte accessor), all sharing the
  existing owned `PPBY.lvec byte` representation, `vmatch_copy_seqbytes`,
  `free_copy_seqbytes` and `seq_vlbytes_conv`.

The generator's per-module `FStar.Bytes`-header emission is now dead under `-pulse`
(`gemstone_uses_fstar_bytes_pulse` returns `false`); the `FStar.Bytes` module headers
are emitted only outside `-pulse`.

### 5. Extraction (KaRaMeL) cleanliness

Spec-only artifacts are marked `noextract` under `-pulse` so the extracted C contains
only the runtime (owned, `Vec`-based) representation and no dead spec types:

- high struct/sum/alias/list/array types and their synth/tag coercions whose runtime
  role is played by the owned low representation,
- generated enum lookup tables and `<n>_mid_tag` ghost helpers.

Leaf-readable types — whose copyful "mid" *is* the high type and is therefore genuinely
used at runtime — are deliberately **kept extractable**. This eliminates
`Prims_list`-style cons-list leakage and dead `krmlinit_globals()` initializers from
the generated C.

The `noextract` markings on the synth coercions and `make_sum'`/`make_dsum'` sum
descriptors are gated on `-pulse` (`if !emit_pulse then "noextract " else ""`): under
the default/`-low`/`-high` backends these stay extractable (the high-backend
`parser32`/`serializer32`/`size32` consume them), so non-Pulse output is unchanged.

## Non-Pulse behavior preserved

All `-pulse`-specific generation is guarded by the `emit_pulse` flag. The default,
`-low`, and `-high` outputs are **byte-identical** to `upstream/master` (verified by
regenerating the full unit + bitcoin grammars with both `qd` binaries and diffing).
In particular, outside `-pulse` the SLow `size32`/`LSZ` emission, `FStar.Mul` operator
qualification, `friend` declarations, the `FStar.Bytes`-based if-then-else backend, and
all enum/type `noextract` markings are exactly as before.

## Testing

- **New `-pulse` unit harness** under `share/everparse/tests/qd/unit`: regenerates from
  `unittests.rfc` + `bitcoin.rfc` via `qd -pulse`, verifies in F\*, extracts to C,
  compiles, and runs round-trip tests.
- Added `t60` / `t61` (little-endian leaf arrays) to `unittests.rfc` for
  little-endian copyful array coverage.
- `make -j16 quackyducky-test` runs **both** the existing Low\* harness (`tests/`) and
  the new Pulse harness (`share/everparse/tests/qd`) and passes end to end
  (verification → C extraction → compile → run).

## Files changed

- **`src/qd/`** — `rfc_fstar_compiler.ml` (the bulk of the `-pulse` generator and the
  split if-then-else codegen), plus `rfc_parser.mly`, `rfc_ast.ml`, `rfc_pretty.ml`
  for the if-then-else grammar (custom constructor names) and AST.
- **`src/lowparse/pulse/`** — the copyful combinator layer: 5 new modules and large
  extensions to ~20 existing `LowParse.Pulse.*` / `LowParse.PulseParse.*` modules,
  including the Seq-native variable-length byte layer in `LowParse.Pulse.SeqBytes`
  (explicit-`l` `_bounded_seq_vlbytes'` and generic-header `_bounded_seq_vlgenbytes`
  `validate` / `jump` / `copyful_parse` / `l2r_safe_writer` / `l2r_safe_size`).
- **`src/lowparse/`** — spec additions in `LowParse.Spec.SeqBytes.fst` (the Seq-native
  `_bounded_seq_vlbytes_gen` and `_bounded_seq_vlgenbytes` parser/serializer/length/
  `_bytes_eq` families) and `LowParse.Spec.IfThenElse.fst`.
- **`share/everparse/tests/qd/unit/unittests.rfc`** — added test types.

## Notes

- `if-then-else` types are a `-pulse`-only feature; requesting the SLow/Low\* backend
  for a grammar that uses them falls back to the legacy `FStar.Bytes`-based codegen
  (unchanged from `upstream/master`).
- The `-pulse` backend is mutually exclusive with `-low`.


tahina-pro:
### 6. No `FStar.SizeT.fits_u64` assumption in `-pulse` output

The `-pulse` backend previously emitted `assume val fits_u64_squash : squash
FStar.SizeT.fits_u64` into **every** generated module and threaded it as a precondition
through the entire validator / jumper / serializer / size combinator API. That
assumption silently bakes in the platform fact "`size_t` is at least 64 bits", which C
only mandates to be **≥ 16 bits** — a soundness gap in the verified-to-C story. It is now
**completely removed**: generated `qd -pulse` parsers, validators and serializers are
sound on **any ≥ 16-bit `size_t`**, with zero `fits_u64` assumptions or preconditions
(`grep fits_u64` over the generated `-pulse` `.fst` files returns nothing).

The techniques, all sound and portable (no static width axiom, no `admit`/`assume`):

- **Portable base-2¹⁵ comparisons** — new module `LowParse.Pulse.SizeComparison`.
  `32768 = 2¹⁵` is the largest power of two guaranteed to fit any ≥16-bit `size_t`
  (`FStar.SizeT.fits_at_least_16`); four divisions by `32768` compute `b / 2⁶⁰`, deciding
  `u64 ↔ size_t` comparisons (`u64_lte_sizet`, `u32_lte_sizet`, `sizet_lte_u32`,
  `sizet_fits_u64`, `sizet_sum_within_budget`) with no platform axiom. In C these compile
  to shifts/masks and are correct at 16/32/64-bit width.
- **`fits` recovery, not assumption** — a wire-read length/count is compared against the
  remaining input (a `size_t`) *before* conversion; `FStar.SizeT.fits_lte` then recovers
  `SZ.fits` and the (already-weakened) `uint32/64_to_sizet` narrows exactly. Over-size
  inputs that cannot fit `size_t` are gracefully rejected — sound, and *more* correct on
  narrow targets.
- **Output-length budget for writers** — serializers recover `fits` from the output
  slice length (`sizet_sum_within_budget (S.len out) …`), so writing succeeds at full
  capacity. Standalone size functions use a portable overflow-checked `size_add_checked`
  (overflow now detected at the `size_t` boundary — `l2r_safe_size_postcond` uses
  `SZ.fits len` instead of `len < pow2 64` — with a `< 2¹⁶` completeness floor, the honest
  fully-portable limit; a `size_add_checked_budget` variant recovers full capacity when a
  budget witness is available).
- **`U32` bounds instead of `size_t` bounds** — the vlbytes/vldata/vlgen/vclist writer
  and size combinators now take their static bounds as `U32.t` and compare the payload
  `size_t` length against them portably, so the old `mk_seq_sizet` helper (which needed
  `fits_u64` to turn a bound ≥ 2¹⁶ into a `size_t`) is deleted.
- **Tightened leaf-size bounds** — pure leaf-size combinators (`leaf_size_pair`,
  `leaf_size_sum`/`_dsum`, `SizeLeaf`) require `parser_kind_high < pow2 16` (discharged
  automatically by `fits_at_least_16`); vestigial `fits_u64` parameters were dropped.
- **`parser_kind_low > 0` for count-prefixed lists** — `validate/jump/copyful_parse_vclist`
  now require it (consistent with the vclist writer/size), the only sound choice for
  representing a wire count as a `size_t` on a narrow platform.

This mirrors the analogous `SZ.fits_u64` elimination done for the CBOR/CDDL Pulse code
(EverParse PR #296). It relies on the already-present `FStar.SizeT` weakening of
`uint32/64_to_sizet` (`requires fits_u32/64 \/ fits (v x)`) and needs no F\* or
Pulse-stdlib change.

tahina-pro:
**Machine integers only (no KaRaMeL Warning 15).** The generated `-pulse` serializers
are free of mathematical (`Prims.int`/`nat`) integers, so the extracted C uses native
machine integers with no bignum/GC dependency. Two combinators that previously leaked
`nat` into extraction were fixed: `LowParse.PulseParse.BCVLI.bounded_bcvli_size` (whose
`nat` bounds are spec-only) is now `inline_for_extraction` so those parameters erase; and
`l2r_safe_writer_bounded_vldata_payload` now receives its header length as a precomputed
machine-integer parameter (`l_sz : SZ.t`, emitted as a `(log256 max)` literal) instead of
computing `log256' max` / `FStar.UInt32.uint_to_t` at runtime — mirroring the already-clean
`_strong` variant. KaRaMeL extraction of the `-pulse` test now reports **zero Warning 15**.

PR #310, opened 2026-08-06
Bump to latest F*

gebner:
This adds a devcontainer for everparse, sorry if I missed an existing one.

This might also be the end of the road for F* 1/2 source file compatibility.  The new introduce/eliminate syntax doesn't work in F* 1 (neither does the interleaving probably).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -23,6 +23,9 @@ RUN apt-get update \
       tmux \
       rustup \
       dotnet-sdk-8.0 \
+      python3-venv \
+      clang \
+      cmake \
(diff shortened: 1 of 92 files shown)


tahina-pro:
Thanks a lot Gabriel!

I did not intend to keep full character-to-character compatibility between the two branches of EverParse. For instance, `fstar2` already had to cope with `FStar.Mul` gone. But such merely syntactic changes are totally fine.

Only for deeper proof changes on `fstar2`, I prefer to backport them to `master` as much as possible.  This way, I can maintain a small enough diff between `master` and `fstar2`.

PR #318, opened 2026-08-28
Port 3D from Low\* to Pulse (`--pulse`)

tahina-pro:
This branch adds a second, complete code-generation backend to 3D, built on
Pulse instead of Low\*, selected by a new `--pulse` flag. The Low\* backend is
untouched and remains the default: every change is either additive or guarded
by `Options.get_pulse ()`.

The port is functionally complete. The new runtime has **no admits**, the
generated C compiles under `-Wall -Wextra -Werror`, and a full parallel copy of
the 3D test corpus builds, links and runs against it.

```
299 files changed, ~27,000 insertions
  lib/everparse/3d       11,234   the Pulse combinator library (20 modules)
  share/everparse/tests  10,301   the --pulse test corpus (37 .3d + 18 subdirs)
  src/3d                  4,412   frontend: --pulse emission, wrappers, Z3TestGen
  src/lowparse              607   ArrayPtr readers, read_u64_le
  .github/workflows         377   --pulse coverage in package/nightly/release CI
```

---

### 1. `lib/everparse/3d` — the Pulse combinator library

A Pulse counterpart to `src/3d/prelude`: `Prelude`, `Kinds`, `ErrorCode`,
`InputStream.{Base,Buffer,Extern,Static}`, `CopyBuffer`, `AppCtxt`,
`Actions.{Base,Common}`, `ProbeActions`, `State`, and `Interpreter`.

Three structural departures from Low\*:

**Memory locations and invariants become `extra_state`.** Low\* threads
`loc`/`inv`/`liveness` through every validator and action. Pulse replaces all
of it with a *state dictionary* (`EverParse3d.State`): a keyed product of
resources, where disjointness is a syntactic property of distinct string keys
rather than an SMT obligation. `Pulse.Lib.ForEvery` carries the pointwise
ownership, and the `call` combinators rename keys at call boundaries. Almost
all of the reasoning discharges by normalization instead of Z3.

**One input buffer becomes three arguments.** `input_stream_inst base_t len_t
pos_t`, with `pts_to : base_t -> len_t -> pos_t -> Seq.seq U8.t -> Seq.seq
U8.t -> slprop`. The `buffer` instance uses `Pulse.Lib.ArrayPtr.ptr U8.t`,
`SZ.t` and `ref SZ.t`; `extern` and `static` instantiate the unused ones to
`unit`, and KaRaMeL erases those arguments.

**Backend-conditional actions without a backend flag.** Rather than a flag,
the interpreter takes `option field_ptr_t`, `option field_ptr_after_t` etc.,
and `Action_field_ptr` carries a `squash (Some? ...)`. The frontend supplies
the instance, so `field_ptr` stays buffer-only and `field_ptr_after`
extern/static-only, exactly as in Low\*, with no linking or `--include` tricks.

The library is extracted to `.krml` (`lib/everparse/3d/krml`) and bundled into
the generated C, rather than shipped as a separate C library the way the Low\*
prelude is.

### 2. `src/3d` — the frontend

`--pulse` switches F\* emission, the `--include` path, the KaRaMeL invocation
and the C wrapper. Notable points:

- **Wrappers** keep their existing argument types, with `_Static_assert`s that
  `size_t` is at least as wide as `uint32_t` and `uint64_t`.
- **Error handler macros**: `buffer`, `extern` and `static` each own a
  `[@@CMacro] assume val error_handler_macro`, and the frontend instantiates
  `validate_with_error_handler` accordingly.
- **Frontend checks** reject probes for `extern`/`static`, and reject nested
  probes that would reuse one copy buffer.
- **Z3 test-case generation** works under `--pulse`, in both harness modes,
  including probes. The SMT half is backend-independent — it runs on the
  `InterpreterTarget` AST — and produces byte-identical witnesses and verdicts
  on both backends; only the emitted C harness differs.
- **Generated parser kinds are reduced to literals** under `--pulse`
  (`norm [delta_namespace ["EverParse3d"; "LowParse"]; ...]`). Without this,
  `and_then_kind`/`glb` nests unfold exponentially during extraction:
  `FieldDependence0.3d` went from OOM at 166 GB to 57 ms.

### 3. `share/everparse/tests/3d` — the test corpus

A full copy of `src/3d/tests` adapted to `--pulse`; `src/3d/tests` is
unchanged. 37 top-level `.3d` files (10 negative) plus 18 subdirectories, all
of which generate, verify, extract, compile and run.

### 4. Packaging and CI

`package.sh` ships `lib/everparse/3d` (sources, `.checked` and pre-extracted
`.krml`) plus `lib/pulse`. Every binary-package test job in
`package-{linux,macos,windows}.yml`, `nightly.yml` and `release.yml` gained a
`--pulse` step, as did the cross-platform hash-interop jobs and the Z3
test-generation steps.

---

## Notable fixes along the way

Several of these are pre-existing bugs the port exposed rather than caused:

- **`Z3TestGen` argument swap** — `Main.fst` passed
  `(print_c_initializers, use_ptr)` to `do_test`/`do_diff_test`, which declare
  `(use_ptr, print_c_initializer)`; `do_diff_test` had a second, compensating
  swap. Both are `bool`, so it typechecked silently, and every existing test
  set the two flags to the same value. `a1b6f3df6`.
- **Copy-buffer reuse** — a buffer reused at a second probe site was validated
  from wherever the first probe left it. Fixed with a `reset` method on the
  `copy_buffer` class. No test outside `specialize_*` would have caught it.
- **`make clean` gaps** — `krml/extract.Makefile` listed `clean-extracted` in
  `clean_rules` without defining it; `tests/3d/iter/Makefile` had no `clean` at
  all (its Low\* original has the same gap); and the subdirectory clean loop ran
  without `set -e`, swallowing both. `68909c201`.
- **Makefile dependency bug** — hand-written `.o` rules listed the generated
  wrapper header but not `EverParse.h`. Harmless under Low\*, where it is
  copied at the start of the pipeline, but under `--pulse` KaRaMeL emits it at
  the *end*, so parallel builds raced. `4f409e1cc`.

---

## Remaining work to reach near-parity with Low\*

Nothing below blocks the branch; none of it is a correctness regression
against the Low\* backend.

### Correctness coverage

1. **`validate_t_exact` on a non-buffer stream is untested.** Under `extern`,
   `stream_truncate` returns `(base_t & unit & unit)`, and how KaRaMeL renders
   that struct is unverified — no subdirectory exercises it. *This is the only
   genuine unknown left, as opposed to polish.*
2. **The standalone F# hash checker has never run against `--pulse` output.**
   Low\*'s three `hashchk-*` targets have no Pulse counterpart.

### Feature gaps

3. **Compile-time comments are not extracted.** `act_with_comment` carries a
   TODO at three sites in `Actions.Base.fst`; `T_with_comment` and the
   `validate_filter` strings are dropped, so `--pulse` C is cosmetically poorer
   than Low\*'s. Needs Pulse-side support.
4. **`3d-doc-test` has no `--pulse` snapshot.** `doc/3d-snapshot/` is
   regenerated and diffed for Low\* only. Best done after (3), to avoid
   re-baselining.
5. **The `--pulse` client ABI is undocumented.** `doc/3d-lang.rst` gives only
   the Low\* copy-buffer hooks; under `--pulse` a client must also define
   `EverParseStreamPos` (returning `size_t *`), and `EverParseStreamLen`
   returns `size_t`. The flag itself and the `(base, len, pos)`
   validator/error-handler ABI are undocumented.

### Verification debt

6. **Four `Tot`→`GTot` TODOs** on the erased bijection argument
   (`State.fst:785`, `Actions.Base.fst:{741,1285}`, `.fsti:{343,485}`).
   Erasure hygiene only; needs ghost bijections to exist first.
7. **`Actions.Base.fst:284`** — an `if/else` that would read better as a
   non-terminal early exit with an `ensures`.

### Accepted differences (by design, not oversight)

- **`static` and `extern` are byte-identical at the C level.** KaRaMeL emits
  plain `extern` for every assumed val, so Low\*'s `static inline` prelude trick
  is unreproducible. They remain separate F\* modules only because each owns its
  `error_handler_macro`. Consequently `field_ptr_after` is allowed for *both*,
  where Low\* restricts it to `extern`.
- **Probes are rejected for `--input_stream extern|static`**, as agreed.
- **Z3TestGen reports the current position** to the error handler where Low\*
  reports each field's start position. Verdicts and witnesses agree; only the
  position column in the trace comments differs.
- **Copy buffers carry their own position.** Low\* passes the start position to
  the validator as an argument; in Pulse it lives in the stream, hence the extra
  `EverParseStreamPos` hook and the `reset` method.

### Trust footprint

`lib/everparse/3d` has **no admits**. The deliberate `assume val`s mirror
Low\*'s one-for-one: the input-stream primitives per backend, the probe
functions, the `[@@CMacro] error_handler_macro`s, and the copy-buffer handle
with its three projections. That last one is exactly Low\*'s bargain —
`stream_of` is assumed and is not really a function of its argument; Low\* pays
for it in liveness, Pulse in ownership, and in both the fresh region is
admitted by fiat inside the already-assumed probe functions.

---

## Testing

- `make 3d-pulse-prelude` — verifies the library (20 modules).
- `make 3d-pulse-krml` — extracts it (148 `.krml`).
- `make 3d-pulse-test` — the full corpus: generation, F\* verification, KaRaMeL
  extraction, C compilation under `-Wall -Wextra -Werror`, and execution.
  Wired into `make test`, and parallel-safe (`-j16`, ~11 min from clean).
- Low\* regressions (`src/3d/tests`, including `z3-testgen-probe-test`) were
  re-run and are unaffected.


opt/Makefile
```diff
@@ -24,9 +24,9 @@ cygwin_local_install += --cygwin-local-install
 endif
 
 # Change these if you want to use your own clones
-FStar_repo=https://github.com/FStarLang/FStar
+FStar_repo=https://github.com/tahina-pro/FStar
 karamel_repo=https://github.com/FStarLang/karamel
-pulse_repo=https://github.com/FStarLang/pulse
+pulse_repo=https://github.com/tahina-pro/pulse
```
tahina-pro: This change (582811799095dd9fa838a8b461c94c73bca5eda6) should be reverted, and the corresponding changes in F* and Pulse should be merged to their respective `fstar1` and `everparse-fstar1` branches.

gebner:
# Review of PR #318: 3D Pulse backend

- PR: <https://github.com/project-everest/everparse/pull/318>
- Reviewed head: `8719bb6e1f199ea4e4cd660f11c16920c176ab9a`
- Base: `e1bfead60deba35db696059c78e77aa406135cdc`
- Recommendation: **request changes**.

This review identifies six correctness/build issues. P1 denotes the
field-location regression, which can give callers a pointer beyond the
validated field. P2 denotes the other correctness or supported-configuration
failures. The intentionally different Pulse ABI, omitted generated comments,
and rejection of probes on extern/static streams are not findings.

## Reproduction environment

Use the reviewed revision and its pinned F*, Pulse, and KaRaMeL dependencies.
The compiler and both runtime backends must already be built. The POSIX
commands below assume the repository's usual `opt/` installation layout:

```sh
# Run from the repository root.
export EVERPARSE_HOME="$PWD"
export FSTAR_EXE="$EVERPARSE_HOME/opt/FStar/out/bin/fstar.exe"
export KRML_EXE="$EVERPARSE_HOME/opt/karamel/out/bin/krml"
export PULSE_HOME="$EVERPARSE_HOME/opt/pulse/out"
export WORK="$(mktemp -d)"
export EP="$EVERPARSE_HOME/bin/3d.exe"
```

Save each named source listing into `$WORK`, unless a different location is
specified. Except for the Windows corpus example, generated files and
executables also go there; none of these MWEs requires modifying the compiler
or runtime. Keep `$WORK` until the examples have been inspected. A successful
reproduction of a bug may intentionally return a nonzero exit status.

The Windows-specific example uses the project's configured Cygwin/MinGW
build environment instead of these Linux tool paths. The packaging example
explicitly unsets Pulse environment variables to model a clean 3D-only build.

## 1. [P1] Field-position and field-pointer actions observe the field end

**Locations:** [Actions.Base.fst:243-245][field-action],
[Actions.Base.fst:1939-1940][dependent-action],
[InputStream.Buffer.fst:377-379][field-pointer].

The consuming validator runs before the action. Pulse stores the cursor in
the stream, but the combinators do not retain a separate field-start
position. Consequently, `field_pos`, `field_pos_32`, `field_pos_64`, and
`field_ptr` use the advanced cursor rather than the start of the field to
which the action is attached. This affects ordinary and dependent/refinement
action paths.

### MWE

Save as `$WORK/FieldStart.3d`:

```c
entrypoint
typedef struct _T(mutable PUINT8* ptr, mutable UINT64* pos) {
  UINT8 prefix;
  UINT8 payload[2]
  {:on-success
    var p = field_ptr;
    *ptr = p;
    var n = field_pos_64;
    *pos = n;
    return true;
  };
} T;
```

Save as `$WORK/field-start.c`:

```c
#include "FieldStartWrapper.h"
#include <inttypes.h>
#include <stddef.h>
#include <stdio.h>

void FieldStartEverParseError(char *type, char *field, char *reason) {
  fprintf(stderr, "%s.%s: %s\n", type, field, reason);
}

int main(void) {
  uint8_t input[3] = {0, 1, 2};
  uint8_t *ptr = NULL;
  uint64_t pos = 99;
  if (!FieldStartCheckT(&ptr, &pos, input, sizeof input))
    return 2;
  printf("pointer offset = %td, field position = %" PRIu64 "\n",
         ptr - input, pos);
  return ptr != input + 1 || pos != 1;
}
```

Generate and run with Pulse:

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse --batch \
  --odir "$WORK/field-pulse" "$WORK/FieldStart.3d"
cc -std=c11 -Wall -Wextra -Werror -I "$WORK/field-pulse" \
  "$WORK/field-start.c" \
  "$WORK/field-pulse/FieldStart.c" \
  "$WORK/field-pulse/FieldStartWrapper.c" \
  -o "$WORK/field-pulse/repro"
"$WORK/field-pulse/repro"
```

**Observed:** `pointer offset = 3, field position = 3`, exit status 1.
The parser succeeds, but the pointer is one past the three-byte input rather
than pointing to the two-byte payload.

Run the same driver against the Low* control:

```sh
"$EP" --fstar "$FSTAR_EXE" --batch \
  --odir "$WORK/field-lowstar" "$WORK/FieldStart.3d"
cc -std=c11 -Wall -Wextra -Werror -I "$WORK/field-lowstar" \
  "$WORK/field-start.c" \
  "$WORK/field-lowstar/FieldStart.c" \
  "$WORK/field-lowstar/FieldStartWrapper.c" \
  -o "$WORK/field-lowstar/repro"
"$WORK/field-lowstar/repro"
```

**Observed control / expected behavior:** `pointer offset = 1, field position = 1`,
exit status 0. Both versions of this MWE were generated, compiled, and run.

The existing `doc/GetFieldPtr.3d` illustrates the same issue on a larger field:
the action on `f2[20]`, following `f1[10]`, obtains `Input + 30` instead of
`Input + 10`.

**Suggested correction:** capture the start before consuming the field and
carry it separately into actions. Use that saved position for field-location
actions while retaining the advanced cursor for subsequent parsing. Preserve
the separately specified semantics of `field_ptr_after` rather than changing
all cursor queries indiscriminately.

## 2. [P2] Mixed-state field-pointer setters generate ill-typed F*

**Locations:** [InterpreterTarget.fst:999-1003][setter-emission] and
[Target.fst:1754-1760][setter-type].

Under `--pulse --input_stream extern` or `static`, a field-pointer setter
for an output structure is emitted at `output_state`. When the parser also
uses a separate mutable scalar parameter, its ambient state dictionary
contains both resources. Unlike ordinary external calls, the
`Action_field_ptr_after_with_setter` emission does not weaken the action into
that larger dictionary. A valid combination of the two features consequently
fails F* verification.

### MWE

Save as `$WORK/MixedSetter.3d`:

```c
output
typedef struct _OUT {
  PUINT8 ptr;
} OUT;

entrypoint
typedef struct _MIXED(mutable OUT *out, mutable UINT32 *p) {
  UINT32 x {:on-success
    *p = x;
    var ok = out->ptr = field_ptr_after(1uL);
    return ok;
  };
} MIXED;
```

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/mixed-setter" "$WORK/MixedSetter.3d"
```

**Expected:** generate and verify the parser, retaining both the output object
and the scalar pointer in its state dictionary.

**Observed:** exit status 1 while verifying `MixedSetter.fst`. The error
occurs before C compilation, so no client implementation of `EverParseStream.h`
is needed to reach it. The relevant diagnostic, with names shortened, is:

```text
Error 19: Subtyping check failed
Expected type
  squash (state_dict_weaken_prop
    (state_dict_singleton k__p (pts_to p))
    MixedSetter.ExternalAPI.output_state)
got type unit
```

In this MWE, the setter pins the surrounding action to `output_state`;
the preceding scalar assignment then fails its weakening obligation. The
generated expression contains:

```fstar
Action_seq (Action_assignment k__p p x ())
  (Action_let
    (Action_field_ptr_after_with_setter
      (B.field_ptr_after_with_setter _) ()
      1uL (set_OUT_deref_ptr out))
    (fun ok -> Atomic_action (Action_return ok)))
```

The enclosing parser is annotated with the product of the `"p"` resource
and `output_state`, but the setter's type only contains `output_state`.

For the Low* control:

```sh
"$EP" --fstar "$FSTAR_EXE" \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/mixed-lowstar" "$WORK/MixedSetter.3d"
```

**Observed control:** the same `.3d` input verifies and extracts successfully.
Both commands above were executed. The source-level reproduction uses
`extern`; the `static` backend shares the problematic emission branch.

**Suggested correction:** construct the output-state action and weaken it into
the enclosing state dictionary before composing it with other actions, as is
already done for ordinary external calls.

## 3. [P2] Stream wrappers report cumulative position as parsed length

**Location:** [Target.fst:1336-1344][stream-wrapper].

The Pulse extern/static wrapper queries `EverParseStreamGetPosition(base)`
only after validation, converts that absolute position into `parsedSize`, and
returns it. Reusing a stream therefore includes bytes consumed by previous
wrapper invocations. The same inflated value is passed to
`EverParseHandleError` and `EverParseRetreat`.

The PR's supplied extern stream increments a persistent `consumed` counter
and does not reset it in `EverParseRetreat`. The old wrapper instead starts
each validator invocation at a logical position of zero and reports that
invocation's consumption.

### MWE

Save as `$WORK/Reuse.3d`:

```c
entrypoint
typedef struct _T {
  UINT8 bytes[12];
} T;
```

Save as `$WORK/stream-count.c`:

```c
#include "EverParseStream.h"
#include "ReuseWrapper.h"
#include <inttypes.h>
#include <stdio.h>

void ReuseEverParseError(char *type, char *field, char *reason) {
  fprintf(stderr, "%s.%s: %s\n", type, field, reason);
}

int main(void) {
  uint8_t bytes[24] = {0};
  struct es_cell cell = {bytes, sizeof bytes, NULL};
  struct EVERPARSE_INPUT_STREAM_BASE_s stream = {.head = &cell};
  uint64_t first = ReuseCheckT(0, &stream);
  uint64_t second = ReuseCheckT(0, &stream);
  printf("first = %" PRIu64 ", second = %" PRIu64 "\n", first, second);
  return first != 12 || second != 12;
}
```

This uses the PR's existing stream implementation, not a special mock. The
stream and its single buffer cell are stack-allocated to keep the driver small.

```sh
STREAM_SRC="$EVERPARSE_HOME/share/everparse/tests/3d/extern/src"
"$EP" --fstar "$FSTAR_EXE" --pulse \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/stream-count" "$WORK/Reuse.3d"
cc -std=c11 -D_DEFAULT_SOURCE \
  -I "$WORK/stream-count" -I "$STREAM_SRC" \
  "$WORK/stream-count/Reuse.c" "$WORK/stream-count/ReuseWrapper.c" \
  "$STREAM_SRC/EverParseStream.c" "$WORK/stream-count.c" \
  -o "$WORK/stream-count/repro"
"$WORK/stream-count/repro"
```

**Observed:** `first = 12, second = 24`, exit status 1.
Both parses succeed and consume 12 bytes, but the second result is cumulative.
**Expected:** `first = 12, second = 12`, exit status 0.

The same driver works with the original Low* stream implementation:

```sh
STREAM_SRC="$EVERPARSE_HOME/src/3d/tests/extern/src"
"$EP" --fstar "$FSTAR_EXE" \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/stream-lowstar" "$WORK/Reuse.3d"
cc -std=c11 -D_DEFAULT_SOURCE \
  -I "$WORK/stream-lowstar" -I "$STREAM_SRC" \
  "$WORK/stream-lowstar/Reuse.c" "$WORK/stream-lowstar/ReuseWrapper.c" \
  "$STREAM_SRC/EverParseStream.c" "$WORK/stream-count.c" \
  -o "$WORK/stream-lowstar/repro"
"$WORK/stream-lowstar/repro"
```

**Observed control:** `first = 12, second = 12`, exit status 0.
Both backends were generated, compiled, and run with this driver.

**Suggested correction:** capture the initial stream position and subtract it
from the final position when computing `parsedSize`. Apply this to both
wrapper-generation branches, including the branch that hoists declarations.

## 4. [P2] 3D-only packaging enters the Pulse-copy branch

**Locations:** [package.sh:215-228][package-copy],
[deps.Makefile:21-23][only-3d], and
[release.yml:448-451][release-pulse].

`EVERPARSE_ONLY_3D=1` sets `NO_PULSE := 1` inside `deps.Makefile`. That variable
is not exported to the packaging shell or included in the environment emitted
by the `env` target. The new shell condition `[[ -z "$NO_PULSE" ]]` therefore
enters the Pulse-copy block in a clean 3D-only build. Because Pulse was not
enabled, `PULSE_HOME` is also unset, and the source of the final copy becomes
`/lib/pulse`. The packaging script uses `set -e`, so that missing source
aborts packaging.

### Minimal, non-destructive diagnostic

Run from the repository root:

```sh
env -u NO_PULSE -u PULSE_HOME EVERPARSE_ONLY_3D=1 \
  make -s --no-print-directory -f deps.Makefile env |
  grep -E '^export (EVERPARSE_HOME|NO_PULSE|PULSE_HOME)='
```

**Observed:** only the `EVERPARSE_HOME` assignment is printed; neither
`NO_PULSE` nor `PULSE_HOME` is exported.

The packaging branch can be demonstrated without copying anything:

```sh
env -u NO_PULSE -u PULSE_HOME EVERPARSE_ONLY_3D=1 bash -c '
  if [[ -z "$NO_PULSE" ]]; then
    printf "Pulse-copy branch selected; source=%s/lib/pulse\n" "$PULSE_HOME"
  else
    printf "Pulse-copy branch skipped\n"
  fi
'
```

**Actual:** `Pulse-copy branch selected; source=/lib/pulse`.
**Expected:** the Pulse-copy branch is skipped for a 3D-only package.

The full reproduction, in a disposable clean checkout with the normal
packaging prerequisites, is:

```sh
env -u NO_PULSE -u PULSE_HOME EVERPARSE_ONLY_3D=1 \
  make -f package.Makefile package
```

This full command builds and packages dependencies; it is not a lightweight
diagnostic. The review used the environment/branch diagnostics rather than
performing a fresh full package build.

**Related release failure:** even after correcting the copy guard, the new
Windows `--pulse` release steps run unconditionally when
`only_build_3d: true`. Such a package intentionally lacks the Pulse runtime.
The build step already passes `EVERPARSE_ONLY_3D=1`, but the Pulse test steps
do not have the matching condition.

**Suggested correction:** propagate the effective Pulse-disabled setting to
the packaging shell, or derive it there consistently from
`EVERPARSE_ONLY_3D`. Guard the Windows Pulse release steps and their artifacts
when `only_build_3d` is selected. Correcting only the package-copy condition
does not make that release configuration work end to end.

## 5. [P2] The Pulse corpus enables a POSIX-only checker on Windows

**Locations:** [tests/3d/Makefile:116-119][pulse-subdirs],
[probe_error_handler_macro/GNUmakefile:173-184][checker-rule],
and [Z3TestGen.fst:2582-2591][checker-includes].

The new `pulse_subdirs` list unconditionally includes
`probe_error_handler_macro`. Its default `world` target includes
`checkertest`, which compiles the standalone checker emitted by
`--test_checker`. That C program includes `<sys/mman.h>` and calls `mmap` and
`munmap`. It cannot compile with the native Windows MinGW toolchain used by
the full-CI job. The original Low* corpus explicitly excludes this directory
on Windows; the Pulse copy drops that exclusion.

### MWE

In the configured Windows Cygwin/MinGW build environment, from the repository
root:

```sh
make -C share/everparse/tests/3d/probe_error_handler_macro checkertest \
  EVERPARSE_HOME="$EVERPARSE_HOME" FSTAR_EXE="$FSTAR_EXE" \
  CC=x86_64-w64-mingw32-gcc
```

The minimal failing operation, once that target has generated its sources,
is:

```sh
cd share/everparse/tests/3d/probe_error_handler_macro
x86_64-w64-mingw32-gcc -I obj -c obj.checkertest/testcases.c
```

**Expected:** the platform-specific test is excluded, or a Windows-compatible
checker is generated.

**Actual failure:** the generated checker requires `sys/mman.h`, which is
unavailable for the native Windows target. Merely compiling with Cygwin's
POSIX compiler instead would not reproduce the CI toolchain.

The generated source begins with:

```c
#define __STDC_FORMAT_MACROS
#include <inttypes.h>
#include "Probe.h"
#include <fcntl.h>
#include <sys/mman.h>
#include <sys/stat.h>
```

The checker generation and compilation were exercised on Linux, and the
unconditional test dependency and POSIX calls were inspected. A Windows
compiler was not available in the review environment, so the Windows command
above is a platform-specific reproduction, not a claim of an executed
Windows run.

**Suggested correction:** restore the Windows exclusion for this test, or
port the checker before enabling it. Audit the copied subdirectory list
against the existing Low* platform guards rather than assuming the
top-level `.3d` exclusions also cover subdirectories.

## 6. [P2] No-copy Pulse makefiles select the Low* runtime header

**Locations:** [GenMakefile.fst:47-53][no-copy-includes] and the new wrapper
include at [Target.fst:1620-1622][pulse-wrapper-include].

With `--pulse --makefile gmake --no_copy_everparse_h`, generated object rules
still add `src/3d/prelude/<backend>` to the C include path. The Pulse wrapper
then includes the Low* `EverParse.h` alongside the newly required
`EverParsePulse.h`. Both define error-handler types and helpers, so even a
one-byte parser with no actions or probes fails to compile.

The header-selection change in `Batch.ml` covers copying the correct header
into the output directory, but does not correct these no-copy include paths.

### End-to-end MWE

Save as `$WORK/Minimal.3d`:

```c
entrypoint
typedef struct _T {
  UINT8 x;
} T;
```

Generate the makefile, then ask it to build the wrapper object:

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse \
  --makefile gmake --no_copy_everparse_h \
  --odir "$WORK/no-copy" --makefile_name "$WORK/no-copy.Makefile" \
  "$WORK/Minimal.3d"

make -f "$WORK/no-copy.Makefile" "$WORK/no-copy/MinimalWrapper.o" \
  EVERPARSE_HOME="$EVERPARSE_HOME" \
  EVERPARSE_INPUT_DIR="$WORK" \
  EVERPARSE_OUTPUT_DIR="$WORK/no-copy" \
  EVERPARSE_CMD="$EP --fstar $FSTAR_EXE --pulse --no_copy_everparse_h"
```

**Expected:** the generated object rule uses Pulse's shipped runtime header
without copying it into the output directory.

**Observed:** F* verification and extraction succeed; C compilation fails and
`make` exits with status 2. The compile command contains:

```text
-I <repo>/src/3d -I <repo>/src/3d/prelude/buffer
```

The resulting errors include:

```text
error: redefinition of 'struct EVERPARSE_ERROR_FRAME_s'
error: conflicting types for 'EVERPARSE_ERROR_FRAME'
error: conflicting types for 'EverParseDefaultErrorHandler'
```

The complete generated-makefile reproduction above was executed. A smaller,
file-free demonstration of the resulting header conflict is:

```sh
printf '#include "EverParse.h"\n#include "EverParsePulse.h"\n' |
  cc -fsyntax-only -x c - \
    -I "$EVERPARSE_HOME/src/3d" \
    -I "$EVERPARSE_HOME/src/3d/prelude/buffer"
```

**Suggested correction:** under Pulse, select
`lib/everparse/3d/krml/<backend>` instead of `src/3d/prelude/<backend>` for
no-copy object rules. Keep `src/3d` for the support headers and preserve the
Low* paths when `--pulse` is absent. The shared rule emitter also serves
NMake, so keep its include selection consistent.

## Evidence and limits

| Finding | Reproduction evidence |
| --- | --- |
| 1 | Pulse runtime failure and passing Low* control, using the same `.3d` input and C driver. |
| 2 | Source-level Pulse F* failure and successful Low* verification/extraction. |
| 3 | Pulse runtime failure and passing Low* control with the corresponding shipped stream implementations. |
| 4 | Executed environment and branch diagnostics; full fresh packaging and release workflow not run. |
| 5 | Generated POSIX-only checker and inspected Windows build wiring; no Windows execution. |
| 6 | Executed generated-makefile build, failing at C compilation after successful verification/extraction. |

The runtime review used source-level tracing in addition to these focused
reproductions. It was not an exhaustive audit of the state-dictionary proofs
or of every client-provided extern primitive. This document does not propose
weakening verification or changing intentionally different backend behavior
to make the examples pass.

[field-action]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/lib/everparse/3d/EverParse3d.Actions.Base.fst#L243-L245
[dependent-action]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/lib/everparse/3d/EverParse3d.Actions.Base.fst#L1939-L1940
[field-pointer]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/lib/everparse/3d/EverParse3d.InputStream.Buffer.fst#L377-L379
[setter-emission]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/InterpreterTarget.fst#L999-L1003
[setter-type]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/Target.fst#L1754-L1760
[stream-wrapper]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/Target.fst#L1336-L1344
[package-copy]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/package/package.sh#L215-L228
[only-3d]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/deps.Makefile#L21-L23
[release-pulse]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/.github/workflows/release.yml#L448-L451
[pulse-subdirs]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/share/everparse/tests/3d/Makefile#L116-L119
[checker-rule]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/share/everparse/tests/3d/probe_error_handler_macro/GNUmakefile#L173-L184
[checker-includes]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/Z3TestGen.fst#L2582-L2591
[no-copy-includes]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/GenMakefile.fst#L47-L53
[pulse-wrapper-include]: https://github.com/project-everest/everparse/blob/8719bb6e1f199ea4e4cd660f11c16920c176ab9a/src/3d/Target.fst#L1620-L1622

gebner:
# Second review of PR #318: 3D Pulse backend

- PR: <https://github.com/project-everest/everparse/pull/318>
- Reviewed head: `05d2798d194baa348fc9f35bad49bed97b10e9d7`
- Recommendation: **request changes**.

This report contains four additional correctness issues found after the fixes
to the original review. It does not repeat the six findings in `REVIEW.md`.
All four are P2: observable behavior differences or failures to compile
previously supported input. The intentionally different Pulse ABI and the
already acknowledged difference in error-callback positions are not findings.

| # | Priority | Finding | Observable consequence |
|---|---|---|---|
| 1 | P2 | Field positions remain cumulative when reusing an extern stream | A valid second record is rejected by an action |
| 2 | P2 | Callback codes disagree with generated public error constants | Clients misclassify validation failures |
| 3 | P2 | Aliased mutable arguments require distinct resource keys | A supported type application fails F* verification |
| 4 | P2 | A user parameter named `output_state` collides with compiler state | Valid input produces an ill-typed interface |

## Reproduction environment

Use the reviewed revision and its pinned dependencies:

| Dependency | Commit |
|---|---|
| F* | `7ef26c00acd7db1b8073b3985ae9733dc5ae5cd1` |
| KaRaMeL | `6c5e3d20ddcb78ac089f624dca8c94df333c53bb` |
| Pulse | `ed54f14a50c7e226fd888c9a011c0b7d51a02272` |

The compiler and both runtime backends must be built. The compiler and Pulse
runtime used for this review were refreshed with
`make 3d-exe 3d-pulse-krml -j4`; do not use a stale executable or the previous
Pulse dependency checkout. At this revision, the executable's version string
can still identify the earlier revision because the generated version file
is not automatically refreshed.

The commands below assume Linux, a C compiler, and the usual `opt/` layout:

```sh
# Run from the repository root.
export EVERPARSE_HOME="$PWD"
export FSTAR_EXE="$EVERPARSE_HOME/opt/FStar/out/bin/fstar.exe"
export KRML_EXE="$EVERPARSE_HOME/opt/karamel/out/bin/krml"
export PULSE_HOME="$EVERPARSE_HOME/opt/pulse/out"
export EP="$EVERPARSE_HOME/bin/3d.exe"
export WORK="$(mktemp -d)"
```

Save each named listing into `$WORK`. Generated files and executables also go
there; none of these examples requires changing compiler or runtime sources.
Run the Pulse and Low* commands separately: a successful reproduction of a
bug can intentionally exit nonzero.

The runtime examples below were generated, compiled, and executed with both
backends during the review. The two compiler-failure examples were run through
`--batch`: Pulse failed at the documented verification stage, whereas Low*
completed verification and extraction.

## 1. [P2] Field positions remain cumulative when reusing an extern stream

**Locations:** [Actions.Base.fst:437-444][field-start],
[Actions.Base.fst:2375-2412][field-pos],
[Target.fst:1339-1358][wrapper-size].

The field-start fix now captures the position before consuming a field.
However, `I.get_position` on an extern stream is the cumulative stream
position, and the field-position actions return that captured value without
subtracting the current wrapper invocation's origin.

The wrapper's separate fix subtracts the starting position when returning
the number of consumed bytes. It does not change the positions visible to
actions. Consequently, two otherwise identical records can take different
action branches, or the second can fail validation solely because a previous
record was read from the same stream.

### MWE

Save as `$WORK/StreamPosition.3d`:

```c
entrypoint
typedef struct _T(mutable UINT64 *offset) {
  UINT32 x {:on-success
    var p = field_pos_64;
    *offset = p;
    return (p == 0);
  };
} T;
```

Save as `$WORK/stream-position.c`:

```c
#include "EverParseStream.h"
#include "StreamPositionWrapper.h"
#include <inttypes.h>
#include <stdio.h>

void StreamPositionEverParseError(char *tn, char *fn, char *reason) {
  printf("error %s.%s: %s\n", tn, fn, reason);
}

int main(void) {
  uint8_t bytes[8] = {0};
  struct es_cell cell = {bytes, sizeof bytes, NULL};
  struct EVERPARSE_INPUT_STREAM_BASE_s stream = {.head = &cell};
  uint64_t first = 99, second = 99;
  uint64_t n1 = StreamPositionCheckT(&first, 0, &stream);
  uint64_t n2 = StreamPositionCheckT(&second, 0, &stream);
  printf("positions: %" PRIu64 " %" PRIu64
         "; consumed: %" PRIu64 " %" PRIu64 "\n",
         first, second, n1, n2);
  return first != 0 || second != 0;
}
```

Generate and run using the shipped Pulse extern-stream implementation:

```sh
STREAM_SRC="$EVERPARSE_HOME/share/everparse/tests/3d/extern/src"
"$EP" --fstar "$FSTAR_EXE" --pulse \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/stream-position-pulse" "$WORK/StreamPosition.3d"
cc -std=c11 -D_DEFAULT_SOURCE \
  -I "$WORK/stream-position-pulse" -I "$STREAM_SRC" \
  "$WORK/stream-position-pulse/StreamPosition.c" \
  "$WORK/stream-position-pulse/StreamPositionWrapper.c" \
  "$STREAM_SRC/EverParseStream.c" "$WORK/stream-position.c" \
  -o "$WORK/stream-position-pulse/repro"
"$WORK/stream-position-pulse/repro"
```

**Observed:** the second invocation reports an action failure, followed by:

```text
positions: 0 4; consumed: 4 4
```

The driver exits 1. Both invocations consume four bytes, but the second action
sees position 4 and returns false. A consumed-byte count of 4 is not evidence
that this action succeeded.

Use the corresponding Low* stream implementation for the control:

```sh
STREAM_SRC="$EVERPARSE_HOME/src/3d/tests/extern/src"
"$EP" --fstar "$FSTAR_EXE" \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/stream-position-lowstar" "$WORK/StreamPosition.3d"
cc -std=c11 -D_DEFAULT_SOURCE \
  -I "$WORK/stream-position-lowstar" -I "$STREAM_SRC" \
  "$WORK/stream-position-lowstar/StreamPosition.c" \
  "$WORK/stream-position-lowstar/StreamPositionWrapper.c" \
  "$STREAM_SRC/EverParseStream.c" "$WORK/stream-position.c" \
  -o "$WORK/stream-position-lowstar/repro"
"$WORK/stream-position-lowstar/repro"
```

**Observed control / expected behavior:**

```text
positions: 0 0; consumed: 4 4
```

The driver exits 0 and neither action fails.

**Suggested correction:** retain the wrapper invocation's origin and account
for it in field-position actions, not just in the returned byte count.
Preserve the intended enclosing-input-relative offsets at ordinary nested
type calls; resetting the origin at every nested validator would introduce
a different regression. The demonstrated runtime failure uses an extern
stream; static-stream position handling should be considered with the same
origin invariant.

This is not the acknowledged difference in the position supplied to error
callbacks: it changes parser acceptance through an ordinary action.

## 2. [P2] Callback error codes disagree with generated public constants

**Locations:** [EverParse3d.ErrorCode.fst:7-26][internal-codes],
[Target.fst:865-872][public-codes],
[Target.fst:995-1056][callback-bridge].

The Pulse runtime assigns different numeric values to several errors, but
the generated wrapper header still emits the Low* public constants. The
default callback bridge casts the code rather than translating it.

| Failure | Pulse runtime code | Generated public constant |
|---|---:|---|
| Action failed | 1 | `EVERPARSE_ERROR_ACTION_FAILED = 5` |
| Constraint failed | 5 | `EVERPARSE_ERROR_CONSTRAINT_FAILED = 6` |
| Unexpected padding | 6 | `EVERPARSE_ERROR_UNEXPECTED_PADDING = 7` |

A client branching on the public constants therefore classifies a constraint
failure as an action failure, and an action failure as a generic error.
The wrong numeric meaning also reaches the extern wrapper's
`EverParseHandleError`; widening the code to `uint64_t` does not fix it.

### MWE

Save as `$WORK/ErrorCodes.3d`:

```c
entrypoint
typedef struct _Constraint {
  UINT8 x { x == 1 };
} Constraint;

entrypoint
typedef struct _Action {
  UINT8 x {:on-success return false; };
} Action;
```

Save as `$WORK/error-codes.c`. The wrapper header supplies the public constants;
the driver calls the low-level validators to record the raw callback codes.

```c
#include "ErrorCodes.h"
#include "ErrorCodesWrapper.h"
#include <stdio.h>

static unsigned observed;
static void handler(const char *tn, const char *fn, const char *reason,
                    uint8_t code, uint8_t *ctxt, uint8_t *base,
                    size_t len, size_t *pos) {
  (void)tn; (void)fn; (void)ctxt; (void)base; (void)len; (void)pos;
  observed = code;
  printf("%s: callback code=%u\n", reason, code);
}

int main(void) {
  uint8_t input[] = {0}, ctxt = 0;
  size_t pos = 0;
  ErrorCodesValidateConstraint(&ctxt, handler, input, sizeof input, &pos);
  printf("constraint macro=%u observed=%u\n",
         (unsigned)EVERPARSE_ERROR_CONSTRAINT_FAILED, observed);
  pos = 0;
  ErrorCodesValidateAction(&ctxt, handler, input, sizeof input, &pos);
  printf("action macro=%u observed=%u\n",
         (unsigned)EVERPARSE_ERROR_ACTION_FAILED, observed);
}
```

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse --batch \
  --odir "$WORK/errors-pulse" "$WORK/ErrorCodes.3d"
cc -std=c11 -I "$WORK/errors-pulse" \
  "$WORK/error-codes.c" "$WORK/errors-pulse/ErrorCodes.c" \
  -o "$WORK/errors-pulse/repro"
"$WORK/errors-pulse/repro"
```

**Observed:**

```text
constraint failed: callback code=5
constraint macro=6 observed=5
action failed: callback code=1
action macro=5 observed=1
```

For the Low* control, save this ABI-adjusted driver as
`$WORK/error-codes-lowstar.c`:

```c
#include "ErrorCodes.h"
#include "ErrorCodesWrapper.h"
#include <inttypes.h>
#include <stdio.h>

static uint64_t observed;
static void handler(const char *tn, const char *fn, const char *reason,
                    uint64_t code, uint8_t *ctxt, uint8_t *base, uint64_t pos) {
  (void)tn; (void)fn; (void)ctxt; (void)base; (void)pos;
  observed = code;
  printf("%s: callback code=%" PRIu64 "\n", reason, code);
}

int main(void) {
  uint8_t input[] = {0}, ctxt = 0;
  ErrorCodesValidateConstraint(&ctxt, handler, input, sizeof input, 0);
  printf("constraint macro=%u observed=%" PRIu64 "\n",
         (unsigned)EVERPARSE_ERROR_CONSTRAINT_FAILED, observed);
  ErrorCodesValidateAction(&ctxt, handler, input, sizeof input, 0);
  printf("action macro=%u observed=%" PRIu64 "\n",
         (unsigned)EVERPARSE_ERROR_ACTION_FAILED, observed);
}
```

```sh
"$EP" --fstar "$FSTAR_EXE" --batch \
  --odir "$WORK/errors-lowstar" "$WORK/ErrorCodes.3d"
cc -std=c11 -I "$WORK/errors-lowstar" \
  "$WORK/error-codes-lowstar.c" "$WORK/errors-lowstar/ErrorCodes.c" \
  -o "$WORK/errors-lowstar/repro"
"$WORK/errors-lowstar/repro"
```

**Observed control / expected consistency:**

```text
constraint failed: callback code=6
constraint macro=6 observed=6
action failed: callback code=5
action macro=5 observed=5
```

These drivers print the mismatch rather than exiting nonzero. The action and
constraint cases were exercised at runtime; the padding mismatch follows
directly from the runtime definitions and emitted constants.

**Suggested correction:** synchronize the emitted public constants with the
codes actually delivered, or translate internal codes consistently at the
public boundary. Do not blindly renumber the internal codes: Pulse validator
postconditions rely on comparisons such as
`res > validator_error_action_failed` to distinguish parse failure from
action failure.

## 3. [P2] Aliased mutable arguments create an impossible key obligation

**Locations:** [InterpreterTarget.fst:884-902][distinct-keys],
[InterpreterTarget.fst:920-926][forward-keys].

The generated callee requires pairwise distinct keys for its mutable
resources. A type application forwards each actual argument's resource key.
When the same pointer is passed to two used formal parameters, those two
mechanisms require a key to be distinct from itself.

### MWE

Save as `$WORK/AliasedArgs.3d`:

```c
typedef struct _WRITE(mutable UINT32 *a, mutable UINT32 *b) {
  UINT32 first {:on-success
    *a = first;
    return true;
  };
  UINT32 second {:on-success
    *b = second;
    return true;
  };
} WRITE;

entrypoint
typedef struct _ROOT(mutable UINT32 *p) {
  UINT8 prefix;
  WRITE(p, p) value;
} ROOT;
```

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse --batch \
  --odir "$WORK/aliased-pulse" "$WORK/AliasedArgs.3d"
```

**Observed:** exit status 1, with Error 19 (`Assertion failed`) while verifying
`AliasedArgs.fst`.

The generated `WRITE` requires:

```fstar
sq_keys_: squash (reveal k__a =!= reveal k__b)
```

The generated `ROOT` calls it with:

```fstar
dtyp__WRITE k__p k__p p p () _ ()
```

The resulting obligation is `reveal k__p =!= reveal k__p`. This is not a
solver timeout that can be fixed by increasing proof resources.

Low* control:

```sh
"$EP" --fstar "$FSTAR_EXE" --batch \
  --odir "$WORK/aliased-lowstar" "$WORK/AliasedArgs.3d"
```

**Observed control / expected behavior:** verification and extraction succeed.
The two writes are sequential; both update the same actual pointer.

**Suggested correction:** accommodate aliased actual resources when
instantiating the callee, for example through alias-aware specialization or
coalescing of resource entries. The caller already deduplicates `p` to a single
dictionary leaf, but that alone does not make the callee's distinct-formal
dictionary usable. Simply removing a proof obligation would not establish the
separating ownership required by the current dictionary representation.

## 4. [P2] A parameter named `output_state` collides with generated state

**Location:** [InterpreterTarget.fst:839-854][state-leaves].

The output resource uses the literal string `"output_state"` both as its
deduplication identity and as its generated dictionary expression. Pointer
resources use the printed user expression as their identity. A valid pointer
parameter named `output_state` therefore collides with the output resource,
and its generated binder also shadows the unqualified dictionary reference.

### MWE

Save as `$WORK/StateNameCollision.3d`:

```c
output
typedef struct _OUT {
  UINT32 value;
} OUT;

entrypoint
typedef struct _T(mutable OUT *out, mutable UINT32 *output_state) {
  UINT32 x {:on-success
    *output_state = x;
    out->value = x;
    return true;
  };
} T;
```

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse --batch \
  --odir "$WORK/collision-pulse" "$WORK/StateNameCollision.3d"
```

**Observed:** exit status 1 while verifying `StateNameCollision.fsti`:

```text
Error 189:
Expected expression of type EverParse3d.State.state_dict
got expression output_state
of type Pulse.Lib.Reference.ref EverParse3d.Prelude.___UINT32
```

The generated declaration binds `output_state` as the scalar reference, then
uses that name where the state dictionary is required. Independently, the
pointer resource and output resource have already been deduplicated under
the same string identity, dropping one resource.

Low* control:

```sh
"$EP" --fstar "$FSTAR_EXE" --batch \
  --odir "$WORK/collision-lowstar" "$WORK/StateNameCollision.3d"
```

**Observed control / expected behavior:** verification and extraction succeed.

**Suggested correction:** qualify or otherwise make the generated dictionary
reference hygienic, and give compiler-owned resource identities a distinct
tag or namespace from user expressions. Both are necessary: qualification
alone leaves the erroneous deduplication in place.

[field-start]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/lib/everparse/3d/EverParse3d.Actions.Base.fst#L437-L444
[field-pos]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/lib/everparse/3d/EverParse3d.Actions.Base.fst#L2375-L2412
[wrapper-size]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/Target.fst#L1339-L1358
[internal-codes]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/lib/everparse/3d/EverParse3d.ErrorCode.fst#L7-L26
[public-codes]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/Target.fst#L865-L872
[callback-bridge]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/Target.fst#L995-L1056
[distinct-keys]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/InterpreterTarget.fst#L884-L902
[forward-keys]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/InterpreterTarget.fst#L920-L926
[state-leaves]: https://github.com/project-everest/everparse/blob/05d2798d194baa348fc9f35bad49bed97b10e9d7/src/3d/InterpreterTarget.fst#L839-L854

tahina-pro:
Thanks Gabriel for the new review!

> 2. [P2] Callback error codes disagree with generated public constants

It is partly intended that the values of the generated public constants in Low* are different from those in Pulse: I wanted to reassign 1 for action failures to do a shortcut in the correctness statements of validators using `> 1` . But yes, I should check that the generated wrapper uses the right values.

> 3. [P2] Aliased mutable arguments create an impossible key obligation

I believe this discrepancy between Low* and Pulse is intended: in the future, we would like to add support for functional correctness of actions (right now we only have memory safety), and without disjoint outparameters, it will be hard to state the semantics of actions in pure F*.

It is nice to see Pulse properly catch this error: this is evidence that my "state dictionary" model is sound.

tahina-pro:
# Response to the first review

Thanks for the detailed review and for the reproducible MWEs — they made this
straightforward to act on.

All six findings have been addressed. I re-ran each of your reproductions
against the current branch (aa2445972); findings 1, 2, 3 and 6 now pass, and 4 and 5 are confirmed by the
diagnostics you supplied and by source inspection.

## Summary

| # | Finding | Status | Fix |
| --- | --- | --- | --- |
| 1 | Field-position and field-pointer actions observe the field end | Fixed | 05d2798d1 |
| 2 | Mixed-state field-pointer setters generate ill-typed F* | Fixed | b6d5943c5 |
| 3 | Stream wrappers report cumulative position as parsed length | Fixed | 03f726249 |
| 4 | 3D-only packaging enters the Pulse-copy branch | Fixed | 8bf9257b4, aa2445972 |
| 5 | The Pulse corpus enables a POSIX-only checker on Windows | Fixed | 3b0b323d1 |
| 6 | No-copy Pulse makefiles select the Low* runtime header | Fixed | 7e8388801 |

## 1. [P1] Field-position and field-pointer actions observe the field end

**Fix:** 05d2798d1, "Report the field start position to field_ptr and
field_pos actions".

Your diagnosis was exactly right: the Pulse `action` type only received
`sl_pos`, the stream cursor, and by the time an `on-success` action runs the
field validator has already consumed the field. The Low* prelude passes two
positions (`pos` and `posf`) precisely to keep these apart.

The Pulse `action` type now takes an extra `start_pos` argument, sampled with
`I.get_position` *before* the field validator runs, together with the bound
`get_position` already guarantees so that the `ArrayPtr.split` in the buffer
backend stays provable. Every validator that invokes an action samples the
position up front and passes it down; `action_bind`, `action_weaken`,
`action_call` and `act_with_comment` thread it through unchanged; and
`field_ptr_t` takes it so the buffer backend splits at the field start rather
than at the cursor.

As you asked, `action_field_ptr_after` is untouched and keeps reading the
current position, matching its separately specified semantics (and matching
Low*, which names its other argument `_posBefore`).

Your `FieldStart.3d` MWE now reports:

```text
pointer offset = 1, field position = 1
```

exit status 0, identical to the Low* control.

The dependent/refinement paths you flagged
(`validate_dep_pair_with_refinement_and_action'`,
`validate_dep_pair_with_refinement_and_action_total_zero_parser'`,
`validate_filter_with_action`) are fixed as well. Checked with an additional
MWE that puts refinements on both reporting fields:

```c
entrypoint
typedef struct _D(mutable UINT64 *pos1, mutable UINT64 *pos2) {
  UINT8 pad[5];
  UINT32 tag { tag == 7 } {:on-success
    var i = field_pos_64; *pos1 = i; return true; };
  UINT16 f { f == 3 } {:on-success
    var j = field_pos_64; *pos2 = j; return true; };
  UINT8 rest[2];
} D;
```

Both backends print `pos1=5 pos2=9`, exit status 0.

**Regression test:** `share/everparse/tests/3d/field_ptr/`, which checks
`field_ptr`, `field_pos_64` and `field_pos_32` at runtime against fields at
known non-zero offsets. Before the fix it reported 14, 16 and base+20 where
Low* reports 10, 14 and base+16.

## 2. [P2] Mixed-state field-pointer setters generate ill-typed F*

**Fix:** b6d5943c5, "Weaken the field-pointer setter into the enclosing state
dictionary".

The `Action_field_ptr_after_with_setter` emission now weakens the
`output_state` action into the enclosing state dictionary before composing it
with the other actions, as ordinary external calls already did
(`EverParse3d.Interpreter.fst`, `src/3d/InterpreterTarget.fst`).

Your `MixedSetter.3d` MWE now succeeds:

```sh
"$EP" --fstar "$FSTAR_EXE" --pulse \
  --input_stream extern --input_stream_include EverParseStream.h \
  --batch --odir "$WORK/mixed-setter" "$WORK/MixedSetter.3d"
```

exit status 0 — F* verification and KaRaMeL extraction both complete.

**Regression test:** `share/everparse/tests/3d/static/src/Test.3d` gained a
case combining an output-object field-pointer setter with a separate mutable
scalar parameter, so the `static` branch of the same emission is covered too.

## 3. [P2] Stream wrappers report cumulative position as parsed length

**Fix:** 03f726249, "Report per-call parsed size from the Pulse stream
wrappers".

`src/3d/Target.fst` now samples `EverParseStreamGetPosition(base)` into a
`startPosition` local *before* the validator call and subtracts it from the
final position, so `parsedSize` is per-invocation. This matches what Low*
obtains from `EverParseGetValidatorErrorPos`, its validator being called with
an explicit start position of 0 each time. The corrected value flows into
`EverParseHandleError` and `EverParseRetreat` as well.

As you noted, both wrapper-generation branches needed it, including the one
that hoists declarations; both are fixed.

Your `Reuse.3d` driver now prints:

```text
first = 12, second = 12
```

exit status 0, matching the Low* control.

**Regression test:** `share/everparse/tests/3d/extern/src/main.c` now calls the
wrapper twice on the same stream and asserts both calls report the same parsed
size.

## 4. [P2] 3D-only packaging enters the Pulse-copy branch

**Fix:** 8bf9257b4 and then aa2445972.

8bf9257b4 fixed the immediate breakage the way you suggested: `deps.Makefile`
now echoes `NO_PULSE` from the `env` target, in the same branch where
`PULSE_HOME` is withheld, so the two stay mutually exclusive by construction
and `package.sh` never dereferences an empty `PULSE_HOME`. It also added the
matching half in `.github/workflows/release.yml`: `build-windows` /
`test-windows` are the only jobs without a job-level `!inputs.only_build_3d`
guard, so each of their four `--pulse` steps got its own.

aa2445972 then went further and dropped the
`EVERPARSE_ONLY_3D=1 ⇒ NO_PULSE:=1` implication entirely, so 3D-only packages
now ship the Pulse backend — which is the point of this PR. Your diagnostic now
reports:

```sh
env -u NO_PULSE -u PULSE_HOME EVERPARSE_ONLY_3D=1 \
  make -s --no-print-directory -f deps.Makefile env |
  grep -E '^export (EVERPARSE_HOME|NO_PULSE|PULSE_HOME)='
```

```text
export PULSE_HOME=/.../opt/pulse/out
export EVERPARSE_HOME=/...
```

so the Pulse-copy branch is taken with a real source, which is now the correct
behaviour for a 3D-only package. The `NO_PULSE` echo added by `8bf9257b4`
remains correct for an explicit `NO_PULSE=1` build.

**One residual note, for your judgement:** given that a 3D-only package now
does contain the Pulse runtime, the `if: ${{ !inputs.only_build_3d }}` guards
8bf9257b4 added to the Windows `--pulse` steps are arguably now too
conservative — the runtime ships but nothing exercises it. That said, it is
consistent with `test-linux` and `test-macos`, which are job-guarded off
entirely for a 3D-only release, so a 3D-only release is simply untested on
those jobs either way. Happy to loosen the guards if you would rather see them
run.

## 5. [P2] The Pulse corpus enables a POSIX-only checker on Windows

**Fix:** 3b0b323d1, "Do not run the POSIX-only Pulse tests on Windows".

`share/everparse/tests/3d/Makefile` now mirrors the guard in
`src/3d/tests/Makefile:55-58`: `pulse_subdirs` unconditionally lists the
portable directories, and `specialize_test`, `specialize_test2`,
`specialize_test_error_handler_macro`, `specialize_tagged_union_array` and
`probe_error_handler_macro` are added only under
`ifneq ($(OS),Windows_NT)`. As you pointed out, the copy already guarded
`exclude_modules` for the `.3d` side of the `UlongToPtr` issue; only the
subdirectory list had been missed.

Verified by dumping `pulse_subdirs` under both settings: with `OS=Windows_NT`
the list drops exactly those five and keeps the rest, so this is a no-op for
the Linux and macOS CI.

This takes the exclusion option rather than porting the checker, as you
allowed. The `Z3TestGen`-generated checker itself is still POSIX-only; porting
it is out of scope for this PR and applies equally to the Low* corpus.

## 6. [P2] No-copy Pulse makefiles select the Low* runtime header

**Fix:** 7e8388801, "Select the Pulse runtime header under
`--pulse --no_copy_everparse_h`".

`print_make_rule` in `src/3d/GenMakefile.fst` now takes a `pulse` flag and
selects `$(EVERPARSE_HOME)/lib/everparse/3d/krml/<backend>` instead of
`$(EVERPARSE_HOME)/src/3d/prelude/<backend>` when `--pulse` is set, keeping
`src/3d` for the support headers (`EverParsePulse{,Endianness}.h`,
`EverParseEndianness.h`) and leaving the Low* paths untouched otherwise. Since
this is the shared rule emitter, NMake gets the same selection.

Your end-to-end MWE now builds. The generated compile line is:

```text
cc -I $WORK -I $WORK/no-copy -I <repo>/src/3d \
   -I <repo>/lib/everparse/3d/krml/buffer \
   -c -o $WORK/no-copy/MinimalWrapper.o $WORK/no-copy/MinimalWrapper.c
```

exit status 0, no `EVERPARSE_ERROR_FRAME` redefinition.

## On the non-findings

Agreed on all three: the Pulse ABI difference, the omitted generated comments
(now partly addressed anyway by 661549572, "3d/pulse: extract compile-time
comments"), and the rejection of probes on extern/static streams are all
intentional.

## Verification performed

Each MWE above was generated, compiled and run on Linux against the current
branch, with the Low* control run alongside where you specified one. Findings 4
and 5 were checked with the environment/branch diagnostics and by comparing the
guards against their Low* counterparts, as in your own review.

tahina-pro:
# Response to the second review

Thanks again for the review and for the fully reproducible MWEs.

Findings **1**, **2** and **4** are fixed. Finding **3** is an intended
difference from Low\* and is not being changed; the rationale is below.

I re-ran all four of your reproductions verbatim against the current branch
(head 447de7bbf, i.e. the reviewed 05d2798d1 plus the four commits below),
with the compiler and the Pulse runtime rebuilt.

## Summary

| # | Finding | Status | Fix |
| --- | --- | --- | --- |
| 1 | Field positions remain cumulative when reusing an extern stream | Fixed | 9904ae095 |
| 2 | Callback codes disagree with generated public error constants | Fixed | cc90140c9 |
| 3 | Aliased mutable arguments require distinct resource keys | Intended (won't fix) | — |
| 4 | A user parameter named `output_state` collides with compiler state | Fixed | 5cc5199c3, 447de7bbf (test) |

---

## 1. [P2] Field positions remain cumulative when reusing an extern stream — fixed

**Fix:** 9904ae095, "3d/pulse: make extern field positions relative to the
validation origin".

Your diagnosis was right: `EverParse3d.InputStream.Extern`'s `get_position`
returned the stream's *cumulative* position, whereas Low\* threads an explicit
position pair whose wrapper always starts at 0.

Rather than change `EverParse3d.Actions.Base`, the fix gives the extern/static
backend's `pos_t` a runtime value: the **origin**, i.e. the absolute stream
position at which the current top-level validation began.

- `stream_pts_to base len origin contents v` now means "the stream consumed
  `origin` bytes before `contents` began".
- The assumed `stream_get_position` keeps its C name and is respecified as
  absolute; a new derived `stream_get_relative_position` subtracts the origin,
  and that is what the instance exposes as `get_position`.
- `stream_trunc_pos` returns the origin unchanged, consistent with
  `stream_truncate`, whose postcondition already continues the parent's
  position accounting.

This is exactly the invariant you suggested: `input_stream_inst` already
threads `sl_pos` unchanged through every combinator and every nested type call,
so **enclosing-input-relative offsets at ordinary nested calls are preserved** —
the origin is *not* reset at every nested validator. `EverParse3d.Actions.Base`
needed no change at all.

You also asked that static-stream position handling be considered with the same
invariant. `EverParse3d.InputStream.Static` is a thin re-export of the `extern`
instance (it exists only to own its own `error_handler_macro`), so it inherits
the fix by construction.

**Client C ABI is unchanged.** Every other assumed val takes the origin as
`Ghost.erased`, so the five primitives the client implements
(`EverParseHas`/`Read`/`Peep`/`Skip`/`Empty`) keep their prototypes; the test's
`EverParseStream.c`/`.h` are untouched by the commit. Only the *generated*
validators and the *generated* default error handler gain one `size_t`, which
`Target.fst` supplies from the `startPosition` the wrapper already sampled.

The buffer backend is unaffected: its `pos_t` is already a `ref SZ.t` that the
wrapper initializes to 0.

### Reproduction (your `StreamPosition.3d` + `stream-position.c`, verbatim)

```
positions: 0 0; consumed: 4 4
```

exit status 0 — identical to your Low\* control. Previously `positions: 0 4`
with exit 1.

### Bonus

As you noted at the end of the finding, this also aligns the extern **error
callback** with Low\*: the generated `DefaultErrorHandler` now reports
`EverParseStreamGetPosition(base) - origin` rather than the raw cumulative
position, closing the previously acknowledged divergence.

### Regression test

9904ae095 adds a `FIELDPOS` entrypoint to
`share/everparse/tests/3d/extern`, which validates the *same* stream object
twice and checks that both runs report offset 0:

```
$ make -C share/everparse/tests/3d/extern
...
Reuse: first = 12, second = 12
FieldPos: offsets = 0 0, sizes = 8 8
```

---

## 2. [P2] Callback error codes disagree with generated public constants — fixed

**Fix:** cc90140c9, "3d: emit Pulse error-code constants when --pulse is set".

As I mentioned in my earlier reply, the *internal renumbering* is intentional:
code 1 is reassigned to "action failed" so that validator postconditions can use
the shortcut `res > validator_error_action_failed`. So, per your own
recommendation, I did **not** renumber the internal codes — I synchronized the
emitted public constants with the codes actually delivered.

`Target.fst`'s `error_code_macros` is now a function of `Options.get_pulse ()`.
The Pulse variant mirrors `lib/everparse/3d/EverParse3d.ErrorCode.fst` exactly
and drops `EVERPARSE_ERROR_GENERIC`, which has no Pulse counterpart. The Low\*
string is byte-for-byte unchanged.

This fixes the `EverParseHandleError` half of the finding too: the extern
wrapper now hands the client codes that match the macros the client branches on.

### Reproduction (your `ErrorCodes.3d` + `error-codes.c`, verbatim)

```
constraint failed: callback code=5
constraint macro=5 observed=5
action failed: callback code=1
action macro=1 observed=1
```

Macro and observed value now agree in both cases. Emitted
`ErrorCodesWrapper.h` under `--pulse`:

```c
#define EVERPARSE_SUCCESS 0ul
#define EVERPARSE_ERROR_ACTION_FAILED 1uL
#define EVERPARSE_ERROR_NOT_ENOUGH_DATA 2uL
#define EVERPARSE_ERROR_IMPOSSIBLE 3uL
#define EVERPARSE_ERROR_LIST_SIZE_NOT_MULTIPLE 4uL
#define EVERPARSE_ERROR_CONSTRAINT_FAILED 5uL
#define EVERPARSE_ERROR_UNEXPECTED_PADDING 6uL
#define EVERPARSE_ERROR_PROBE_FAILED 7uL
```

The unexpected-padding entry, which you derived rather than ran, is covered by
the same table.

Low\* is untouched — regenerating the same `.3d` without `--pulse` still emits
`GENERIC 1 / … / ACTION_FAILED 5 / CONSTRAINT_FAILED 6 / UNEXPECTED_PADDING 7`.

### Regression test

The `goto_return` snapshot pins these constants, so it doubles as the
regression test; it is updated in the same commit.

### Note for clients

The numeric values now differ between the two backends *by design*. Portable
clients must branch on the `EVERPARSE_ERROR_*` macros rather than on literals —
which is what those macros are for, and which now works. The Low\*-vs-Pulse
differential test is unaffected: it compares verdicts, out-parameters, copy
buffers and error *reason strings* and field names, not numeric codes.

---

## 3. [P2] Aliased mutable arguments create an impossible key obligation — intended

Confirmed still reproducing on the current branch, with exactly the diagnosis
you give:

```
* Error 19 at .../AliasedArgs.fst(151,2-169,11):
  - Assertion failed
```

from the obligation `reveal k__p =!= reveal k__p`.

I am deliberately keeping this as a difference from Low\*. We would like to add
functional correctness of actions in the future (today the Pulse backend only
guarantees memory safety), and without disjoint out-parameters it is hard to
state the semantics of actions in pure F\*. Separating ownership of the mutable
out-parameters is what makes that statement possible, so the distinct-key
requirement on the state dictionary is a feature rather than an accident.

It is in fact nice to see Pulse catch this: it is evidence that the state
dictionary model is sound. Your own note that "simply removing a proof
obligation would not establish the separating ownership required by the current
dictionary representation" is exactly right.

Acknowledged residual, tracked separately from this PR: the restriction
currently surfaces as an opaque F\* `Error 19` rather than a frontend
diagnostic, and it is not yet written down in `doc/3d-lang.rst`. Both are worth
improving, but neither is a soundness or behavioural issue.

---

## 4. [P2] A parameter named `output_state` collides with generated state — fixed

**Fix:** 5cc5199c3, "3d: avoid name collisions with the generated Pulse output
state", plus the regression test in 447de7bbf.

You correctly identified that *both* halves have to be fixed, and both are:

1. **Hygiene.** The generated definition is renamed to `___output_state`. 3D
   rejects user identifiers starting with `___`, and `print_ident` only ever
   emits that prefix in front of an uppercase letter, so `___output_state` is
   unreachable from user syntax.
2. **Identity.** The state-dictionary key moves into a reserved namespace as
   `"#output"`. `pulse_sanitize_key` maps every character outside
   `[0-9A-Za-z_.']` to `_`, so `#` can never occur in a key derived from a user
   expression.

Doing (2) required separating the *deduplication identity* of a state-dict leaf
from its *key expression*: `pulse_state_dict_leaves` now returns
`(identity, key_expr, dict_expr)` triples, because `pulse_state_dict_keys`
previously re-derived the key by applying `pulse_key_of` to the identity, which
would have mangled `#output` back into `_output`.

### Reproduction (your `StateNameCollision.3d`, verbatim)

`3d --pulse --batch` now succeeds: generation, F\* verification and KaRaMeL
extraction all pass. The generated sources show both halves in effect:

`StateNameCollision.fsti` — the user's binder and the generated dictionary now
coexist:

```fstar
(k__out k__output_state: FStar.Ghost.erased string)
(output_state: (Pulse.Lib.Reference.ref ((___UINT32))))
(sq_keys_: squash ((FStar.Ghost.reveal k__output_state) =!= "#output"))
...
(state_dict_prod (state_dict_singleton (FStar.Ghost.reveal k__output_state)
        (pts_to output_state #1.0R))
    ___output_state)
```

`StateNameCollision.ExternalAPI.fsti` — the output resource keyed on the
reserved name:

```fstar
let ___output_state:EverParse3d.State.state_dict =
  EverParse3d.State.state_dict_singleton "#output" ___output_state_slprop
```

Note the emitted `sq_keys_` obligation is now the *satisfiable*
`k__output_state =!= "#output"`, so the two resources are no longer
deduplicated into one.

### Regression test

447de7bbf adds `share/everparse/tests/3d/output_state_name`, a compile-only
(`--batch`) test whose entrypoint has a mutable output parameter literally named
`output_state`, wired into `pulse_subdirs`. It exercises both halves: the
generated `Test.fst` binds the user's `output_state` alongside references to
`___output_state`, and `Test.ExternalAPI.fsti` keys the generated entry on
`"#output"`. It builds nothing executable, so it is not a pulse-diff candidate.

---

## Verification performed

Reviewed-branch head 447de7bbf. Compiler (`bin/3d.exe`) and the Pulse runtime
(`3d-pulse-krml`) rebuilt from that revision; all four MWEs saved and run
verbatim as given in the review.

| Check | Result |
| --- | --- |
| MWE 1 `StreamPosition` (`--pulse`, extern stream), generate + cc + run | `positions: 0 0; consumed: 4 4`, exit 0 — matches Low\* control |
| MWE 2 `ErrorCodes` (`--pulse`), generate + cc + run | macro == observed for both constraint (5) and action (1) |
| MWE 2, Low\* regeneration | constants byte-identical to before the fix |
| MWE 3 `AliasedArgs` (`--pulse --batch`) | still `Error 19` — intended, see above |
| MWE 4 `StateNameCollision` (`--pulse --batch`) | `EverParse succeeded!` (generation + verification + extraction) |
| `make -C share/everparse/tests/3d/extern` | passes; `FieldPos: offsets = 0 0, sizes = 8 8` |
| `make -C share/everparse/tests/3d/output_state_name` | passes |

